PR #1, opened 2018-05-15
Declare const parameters as const

tstenner:
This PR picks up 322b5244e1faf9d77b42c0f10944523db0d74dab by @cboulay (reverted because the const qualifiers caused the function names to be mangled and therefore not found when using C linkage) and adds the const qualifiers where needed (i.e. before most pointers to read-only buffers) and marks functions in `*_c.cpp` files as `extern "C"` so the names don't get mangled.


tstenner:
CI: [MSVC 2008](https://ci.appveyor.com/project/tstenner/liblsl/build/job/xo8ft4l088q0q2je), [Linux+OS X](https://travis-ci.org/labstreaminglayer/liblsl/builds/378628225)

PR #2, opened 2018-08-17
Only allow explicit template instantiations in one object file

tstenner:
This ~should fix~ fixes the `multiple definition of `typeinfo name for boost::archive::...` errors on ARM


cboulay:
[liblsl32_arm.zip](https://github.com/labstreaminglayer/liblsl/files/2297649/liblsl32_arm.zip)
When merged with master I was able to use the cross-compile instructions from Ubuntu to build this hopefully Raspberry Pi compatible shared object.

I also tested the native build in Ubuntu 18.04 using the Benchmark tools and it seemed to work there. I'll update this post after I've tested in Mac and Windows. (As far as I know we don't have a unit test suite but let me know if I'm wrong)

tstenner:
I could reproduce the build failure with the master branch on my 3B (gcc 6.3), whereas this branch builds without any warnings.

cboulay:
@tstenner Built this on his Pi:
https://github.com/labstreaminglayer/liblsl/releases/download/1.12/liblsl-1.12.0-Linux-ARM7.deb

PR #3, opened 2018-09-14
Lsltypefwds

tstenner:
Motivation: debugging apps is needlessly complicated because the app and lsl are separated by an opaque C API.

So, without much ado, the result of this PR: a debugger that can browse through the outlet implementation class via an lsl::stream_outlet object:

image

The [first commit](https://github.com/labstreaminglayer/liblsl/commit/0b63a40c46a750b74047edf900e0e5ddad1e1ba3) declares the lsl types as typedefs to forward declarations to the real classes (so instead of "This is a pointer to some kind of struct" the compiler knows "This is a pointer to a class named `lsl::stream_info_impl` I might see later on").

Since the types are know to the compiler in the implementation files, all casts between the dummy C interface type and the actual type can be removed (commit 2).


cboulay:
I chatted with @chkothe about this briefly. He has some concerns, but maybe they can be addressed. I've asked him to hop on here and share his thoughts with you.

tstenner:
> However, it seems that the core issue is the goofy way that the header files are setup. The core library is in C++. lsl_c.h wraps the C++ code with C code. Then lsl_cpp.h wraps the C functions declared in lsl_c.h.

For C++ it's weird, but the pattern (the (hourglass interface)[https://github.com/CppCon/CppCon2014/tree/master/Presentations/Hourglass%20Interfaces%20for%20C%2B%2B%20APIs]) is well established and works around the lack of a C++ ABI and the hard to debug binary compatibility issues between minor compiler version / compiler flag differences.

I've thought about the legacy abi, but having it exposed in the shared libraries would cause a lot more mysterious breakages than e.g. the current ones we're having with the mex files. One idea was to have a shared library with the stable, standardized C API and a shared library with the additional C++ API, but the CMake feature to make this easy has just arrived in CMake 3.12 and almost no Linux distribution ships it.

> It would make more sense if lsl_cpp.h refers only to C++ code, and lsl_c.h wraps the C++ code

The only consumer for the C++ API is C++ code, whereas _every_ other language uses the C API / ABI. The only other languages that might share the C++ interface could be Rust, D and R, and I don't see either one of them becoming relevant for LSL apps any time soon.

PR #16, opened 2018-12-13
Use well defined integer sizes

tstenner:
This PR changes user facing integer types to have a well defined size, mostly because types like "long" can be either int32 or int64, depending on OS, compiler and target arch pointer size. It also changes buffer sizes to uint32, because 32 bit clients can't handle more than that and we need 64 and 32 bit clients to be compatible to each other.

This breaks the ABI, so we shouldn't merge this for 1.13.

See also: https://github.com/sccn/lsl_archived/issues/165, https://github.com/sccn/lsl_archived/pull/320

### DIFF
--- a/include/lsl_c.h
+++ b/include/lsl_c.h
@@ -506,7 +506,7 @@ extern LIBLSL_C_API void lsl_destroy_outlet(lsl_outlet out);
  */
 extern LIBLSL_C_API int32_t lsl_push_sample_f(lsl_outlet out, const float *data);
 extern LIBLSL_C_API int32_t lsl_push_sample_d(lsl_outlet out, const double *data);
-extern LIBLSL_C_API int32_t lsl_push_sample_l(lsl_outlet out, const long *data);
+extern LIBLSL_C_API int32_t lsl_push_sample_l(lsl_outlet out, const int64_t *data);
 extern LIBLSL_C_API int32_t lsl_push_sample_i(lsl_outlet out, const int32_t *data);
 extern LIBLSL_C_API int32_t lsl_push_sample_s(lsl_outlet out, const int16_t *data);
 extern LIBLSL_C_API int32_t lsl_push_sample_c(lsl_outlet out, const char *data);
@@ -519,7 +519,7 @@ extern LIBLSL_C_API int32_t lsl_push_sample_v(lsl_outlet out, const void *data);
  */
 extern LIBLSL_C_API int32_t lsl_push_sample_ft(lsl_outlet out, const float *data, double timestamp);
 extern LIBLSL_C_API int32_t lsl_push_sample_dt(lsl_outlet out, const double *data, double timestamp);
-extern LIBLSL_C_API int32_t lsl_push_sample_lt(lsl_outlet out, const long *data, double timestamp);
+extern LIBLSL_C_API int32_t lsl_push_sample_lt(lsl_outlet out, const int64_t *data, double timestamp);
 extern LIBLSL_C_API int32_t lsl_push_sample_it(lsl_outlet out, const int32_t *data, double timestamp);
 extern LIBLSL_C_API int32_t lsl_push_sample_st(lsl_outlet out, const int16_t *data, double timestamp);
 extern LIBLSL_C_API int32_t lsl_push_sample_ct(lsl_outlet out, const char *data, double timestamp);
@@ -534,7 +534,7 @@ extern LIBLSL_C_API int32_t lsl_push_sample_vt(lsl_outlet out, const void *data,
  */
 extern LIBLSL_C_API int32_t lsl_push_sample_ftp(lsl_outlet out, const float *data, double timestamp, int32_t pushthrough);
 extern LIBLSL_C_API int32_t lsl_push_sample_dtp(lsl_outlet out, const double *data, double timestamp, int32_t pushthrough);
-extern LIBLSL_C_API int32_t lsl_push_sample_ltp(lsl_outlet out, const long *data, double timestamp, int32_t pushthrough);
+extern LIBLSL_C_API int32_t lsl_push_sample_ltp(lsl_outlet out, const int64_t *data, double timestamp, int32_t pushthrough);
 extern LIBLSL_C_API int32_t lsl_push_sample_itp(lsl_outlet out, const int32_t *data, double timestamp, int32_t pushthrough);
 extern LIBLSL_C_API int32_t lsl_push_sample_stp(lsl_outlet out, const int16_t *data, double timestamp, int32_t pushthrough);
 extern LIBLSL_C_API int32_t lsl_push_sample_ctp(lsl_outlet out, const char *data, double timestamp, int32_t pushthrough);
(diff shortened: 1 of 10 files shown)


mgrivich:
Can you load the library in pylsl and check the version, regardless of what pull_chunk_l's prototype looks like? Or can python not handle it if any methods change prototypes, even if they are never called?

tstenner:
The FFI library we're using right now can handle changing function signatures, but the python package isn't my main worry.
Due to the licensing problems we can't build Matlab packages and the Matlab users are very enthusiastic about mixing dlls and mex files from who knows where until Matlab doesn't complain anymore.

mgrivich:
Could you explain these licensing issues?

tstenner:
For (at least) C++, Python, Java and C#, it takes two config files to build, test and upload the pylsl bindings for every Python version from 2.7 to 3.7 on macOS, Ubuntu 18.04, Ubuntu 16.04, CentOS 5, Windows 7 32 bit and Windows 7 64 bit.

For Matlab, someone who has a Matlab license builds the bindings, and test them on a local PC that's configured for the daily lab work (i.e. probably no firewall, up to date packages and so on).
So we have Matlab bindings for Windows 10 with Matlab R2018, Matlab R2011 for Windows 7 compiled against an older liblsl and that's it. For most of the Matlab issues, we can't do anything except tell the people to compile the bindings themselves.

(And even if we had Matlab licenses, we would have no way to build the bindings on a CI system because we'd have to somehow upload the 12GB installer, install Matlab, register the install which is probably against the license terms and then build the bindings).

The new function would ensure old bindings could use a newer library version without rebuilding and future versions of the bindings could just require a liblsl version with the new function.

mgrivich:
We really can't support mixing and matching of bindings, dlls, and matlab version. That way lies madness. The only reasonable solution is to provide either clear building instructions or builds or both. I have a matlab license, and am willing to provide builds upon request. I have put one up at https://github.com/labstreaminglayer/liblsl-Matlab/releases/tag/1.13.0-b13.

tstenner:
I agree about not supporting this kind of library chaos, but we can resolve this issue without an ABI break. 

I've added the int64 functions and switched the C++ API header to use them, so we could even ship it with 1.13.

mgrivich:
"I've added the int64 functions and switched the C++ API header to use them, so we could even ship it with 1.13."

Please switch this back to how I have it in my original pull request. I can redo it from master if it has been lost in the repository shuffle. The Java and Matlab wrappers do not even wrap the offending functions. If someone used the C/C++ functions, they'd get obvious data corruption. You'd know more about Python. Regardless, though, I would shocked if someone is using the broken functions. The changes you propose would make the ABI ugly forever, instead of just in the past.

tstenner:
> I can redo it from master if it has been lost in the repository shuffle.
It's still there, e.g. at bdb9841a99bb60cf8056b89b0067877d32673d8b, but it also switches buffer sizes to uint32_t.

> The Java and Matlab wrappers do not even wrap the offending functions
The Matlab wrapper definitively does:

https://github.com/labstreaminglayer/liblsl-Matlab/blob/17c89909f8f28a1cdd96eef4a444432c4ace0753/mex/lsl_loadlib_.c#L185-L187

https://github.com/labstreaminglayer/liblsl-Matlab/blob/17c89909f8f28a1cdd96eef4a444432c4ace0753/mex/lsl_common.h#L103-L105

> If someone used the C/C++ functions, they'd get obvious data corruption.
Only on platforms that define `long` as `int32`, so [all modern 32bit OSes and Windows](https://stackoverflow.com/a/384672/73299).

> Regardless, though, I would shocked if someone is using the broken functions.
As said above, they are only broken on some platforms.

> The changes you propose would make the ABI ugly forever, instead of just in the past.
And I'd take full responsibility for the ugliness (at least until we can reasonably remove the old functions in 4-5 years), but I wouldn't for hard to diagnose segfaults.

mgrivich:
"The Matlab wrapper definitively does:"

I was a bit hasty. For whatever reason, lsl_push_sample_l has an address after lsl_loadlib, but lsl_pull_sample_l does not. See https://github.com/labstreaminglayer/liblsl-Matlab/issues/7#issuecomment-532045933

"And I'd take full responsibility for the ugliness (at least until we can reasonably remove the old functions in 4-5 years), but I wouldn't for hard to diagnose segfaults."

I don't really care about responsibility. I care about keeping the library clean. Segfaults are not that hard to diagnose, for people than experienced in C. And step 1: update the library will fix it in most cases. But if you like, you can refer such user reports to me.

tstenner:
> Segfaults are not that hard to diagnose, for people than experienced in C.

I'm not worried about those. It's easy to recognize the problem when you get the crash report from matlab, scroll down below the system information and dumped registers to find `lsl_loadlib_.mex` in the stacktrace and make the connection that you can get help for `lsl_loadlib` in the liblsl-Matlab bugtracker. I don't expect most academic Matlab users to do this.

ABI breaks would also prevent me from replacing the ancient liblsl binary shipped with certain vendor products with a newer version that includes fixes for the locale bug and the multiple network interfaces on windows (as soon as that's fixed).

If it's about cleanliness, I could replace the `i64` in the function names with the more common `q` for `quadword`. That way, the function names keep the same width and nomenclature.

mgrivich:
"ABI breaks would also prevent me from replacing the ancient liblsl binary shipped with certain vendor products with a newer version that includes fixes for the locale bug and the multiple network interfaces on windows (as soon as that's fixed)."

This is why we decided to put the integer types fix in liblsl 1.14. 

"If it's about cleanliness, I could replace the i64 in the function names with the more common q for quadword. That way, the function names keep the same width and nomenclature."

It's not what the name is, it is that the function names are changed from the original.

samuelpowell:
@cboulay suggested in #12 that I provide an opinion here. I am hesitant since i) I don't know the library very well, ii) I do not have to worry about legacy support, nor the wonderful world of MEX, and iii) I'm unlikely to have to do the work. But on the off-chance that another view is helpful, I would ask if it is not be possible to achieve the two (somewhat orthogonal goals) of fixing the typing and respecting the legacy interface by:

1. Use length defined integer types throughout the library, and provide a new set of C API functions with concrete suffixes for _all_ types (`_i8`,`_i16`, `_i32`, `_i64`, `_f32`, `f_64`, etc...). 

2. Map legacy functions to the new functions using a set of compiler gaurds that work on architecture and word size, e.g., map `_l` functions to `i32` on Windows and on all 32-bit architectures.

3. Deprecate the legacy mappings and remove in the future when you change the major version number.

tstenner:
I'd say you're perfectly qualified to give an opinion since i) you're not prejudiced by too much exposure to the internals ii) (let's strike that one), iii) only interested in the quality of the outcome and iv) closer to being an end-user who actually has to deal with the change.

1. would've been a great idea back then when the functions were named, but now I think it's too extreme (it's breaking all of the functions instead of just one).

2. As much as I prefer the names you're proposing, they aren't just used in the C headers but also in every language binding so we can't redirect them automatically.

3. I interpret that as a +1 for my proposal :-)

samuelpowell:
Thanks @tstenner!

I think I'm missing something regarding point 2. Why can you not just have compiler guarded stubs to maintain the legacy interface, e.g., 

```
#ifdef (32-bits)
LIBLSL_C_API int32_t lsl_push_sample_l(lsl_outlet out, const long *data)
{
  return lsl_push_sample_i32(lsl_outlet out, int32_t *data)
}
#else
.
.
.
```

tstenner:
If we could recompile all the binaries out there and replace all language bindings without a transition period, that would be the ideal solution. Even for the open sourced users we can't do that, and I wouldn't even begin with the others...

Also, the language bindings have the names as strings so we'd need to add checks against someone loading a too old library *and* have runtime checks which function should be used.

tstenner:
Even though I'm still not a fan of your approach, I had some fun with MSVC claiming `long` is definitely a 32 or 64 bit int, but neither `int32_t` nor `int64_t`. So I pushed a commit to simply switch long to int64 in the push/pull functions so we can put this PR and some technical debt to rest.

tstenner:
1.13 is branched and released, so I'm starting the 1.14 development with this important change.
I've also pushed a change (3e3aefcc877dee3ddd5f1cfbdbdcfda055725ee2) to omit the pointer size suffix from library names (#17), so it's less likely that the new version with the incompatible int sizes will be loaded by accident.

PR #19, opened 2019-01-12
Fix timeout for pull_chunk_multiplexed

tstenner:
Before, the `pull_chunk_multiplexed` overload for `std::vector<T>` would wait for `timeout` seconds after the last sample, which would never stop with sampling rates >= 1/timeout.
The new behavior is to retrieve samples until the timeout expires.

### DIFF
--- a/include/lsl_cpp.h
+++ b/include/lsl_cpp.h
@@ -991,6 +991,8 @@ namespace lsl {
 		 * @param chunk A vector to hold the multiplexed (Sample 1 Channel 1,
 		 * S1C2, S2C1, S2C2, S3C1, S3C2, ...) samples
 		 * @param timestamps A vector to hold the timestamps or nullptr
+		 * @param timeout Time to wait for the first sample. The default value of 0.0 will not wait
+		 * for data to arrive, pulling only samples already received.
 		 * @param append (True:) Append data or (false:) clear them first
 		 * @return True if some data was obtained.
 		 * @throws lost_error (if the stream source has been lost).
@@ -1003,12 +1005,16 @@ namespace lsl {
 			chunk.clear();
 			if(timestamps) timestamps->clear();
 			}
+			std::vector<T> sample;
+			double ts;
+			if ((ts = pull_sample(sample, timeout)) == 0.0) return false;
+			chunk.insert(chunk.end(), sample.begin(), sample.end());
+			if (timestamps) timestamps->push_back(ts);
 			const auto target = samples_available();
 			chunk.reserve(chunk.size() + target * this->channel_count);
 			if(timestamps)
 				timestamps->reserve(timestamps->size() + target);
-			std::vector<T> sample;
-			while (double ts = pull_sample(sample, timeout)) {
+			while ((ts = pull_sample(sample, 0.0)) != 0.0) {
 #if __cplusplus > 199711L || _MSC_VER >= 1900
 				chunk.insert(chunk.end(),
 					std::make_move_iterator(sample.begin()),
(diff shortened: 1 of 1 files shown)


cboulay:
I'm reluctant to make such an important change. While I agree that the current function is broken, essentially what you are doing here is changing the `timeout` argument to something more akin to `poll_time`.

My intuition for 'timeout' would be that any >0 timeout would return as soon as samples are available, including 'right away' if there are samples waiting. With this commit, it seems to be adding this much time to 'now' and returning however many samples are available after that time.

I think the better solution would be to make it behave as a real timeout.

cboulay:
When making changes to the core library, please keep the changes to a minimum so they are easier to evaluate. Most of the changed lines in the PR are just formatting changes and they are obscuring the matter at hand.

Using int16 as an example, I'm just doing a little dive through the code to make sure the meaning of `timeout` is consistent.

1. [`pull_chunk_multiplexed` in lsl_cpp.h](https://github.com/sccn/liblsl/blob/966d0b36702594bc147d3aaa9aeb4bcddc2b07ea/include/lsl_cpp.h#L989-L1022)
1. [`pull_sample_s` in lsl_inlet_c.cpp](https://github.com/sccn/liblsl/blob/d60e3d9f2c1f0820b65b7d8eecbe2bf4af739309/src/lsl_inlet_c.cpp#L221-L223)
1. [`pull_sample_noexcept` in stream_inlet_impl.h](https://github.com/sccn/liblsl/blob/5281edbf8a8f05c802c83746bb9400042eaf1160/src/stream_inlet_impl.h#L98-L115). Notice that there is no docstring for `timeout`. There is only an outdated docstring for `block` which no longer exists.
1. [`pull_sample_typed` in data_receiver.h](https://github.com/sccn/liblsl/blob/37d26cf120d199e03fa77fb081c9a31dc6d9cf4e/src/data_receiver.h#L53-L72).
1. [`consumer_queue::pop_sample(double timeout)`](https://github.com/sccn/liblsl/blob/51e88c4327eb189ccac8d781f31f4772c74821c2/src/consumer_queue.cpp#L49-L65).

That looks good except for how it's used at the top in `pull_chunk_multiplexed`. Indeed your change is the more-correct version of how it works currently, but unfortunately neither version is consistent with the common definition of `timeout`.

The following change to `pull_chunk_multiplexed` can return early if no samples remain, and it won't push the timeout into the future.

 `while (double ts = pull_sample(sample, timeout)) {`
-->
```
double clock_timeout = timeout + lsl_local_clock();
while (double ts = pull_sample(sample, clock_timeout - lsl_local_clock())) {
```

Maybe tangential to this discussion, what happens if the buffer has more samples than was reserved in `chunk`?

tstenner:
> When making changes to the core library, please keep the changes to a minimum so they are easier to evaluate.

Done, clang-tidy was a bit overzealous.

> ```double clock_timeout = timeout + lsl_local_clock();```
> ```while (double ts = pull_sample(sample, clock_timeout - lsl_local_clock())) {```

That'd work, but pull_chunk is supposed to be the faster alternative to pull_sample and pull_sample + one call to lsl_local_clock will be slower than this.

> Maybe tangential to this discussion, what happens if the buffer has more samples than was reserved in chunk?

Then chunk will be resized automatically, which will be unfortunate for performance but reserve is only meant as a hint. We can't stop once it's full because samples_available isn't required to actually tell how many samples are there, only if there are none (0) or any (>=1).

If pushed an idea I've had this morning: pull the first sample with the specified timeout and pull all samples that can be pulled without blocking afterwards. That way we get the timeout behaviour of pull_sample for free and the relatively overhead free pulling afterwards.

cboulay:
Can you please rebase against master so it will trigger the CI builds?

tstenner:
The CI is happy, should I merge it and tag the commit as beta 4 so we can upload a new prerelease?

tstenner:
Done.

PR #22, opened 2019-04-24
Build manylinux binary with docker

tstenner:
See labstreaminglayer/liblsl-Python#6


cboulay:
Go ahead and merge. If you then tag it with a new version number, we might see it deploy to the Release page correctly.

Issue #24, opened 2019-03-21
floating point conversion is locale dependent


sheinke:
Hi there,
I have encountered a rather persistent issue with the liblsl-library on Ubuntu 16.04 LTS, compiling with GCC 5.4.0:
While other hosts in the network (including the localhost) can **detect** outgoing streams, there seems to be a problem with the input parsing: Every field of the stream_info object that is returned by the lsl::resolve_streams function is empty (except for IPv4 and IPv6 addresses). My Ubuntu machine can receive data perfectly fine, this issue seems to concern outgoing connections only. This behavior is not influenced by the receiving machine, I get the same error on Ubuntu and Windows 10. When trying to build a stream_inlet with this faulty stream_info, it says:

*Note: The stream named '(invalid: bad lexical cast: source type value could not be interpreted as target)' could not be recovered automatically if its provider crashed because it does not specify a unique data source ID.*

The part mentioning the failed cast seems to originate in the boost-backend.

This is an excerpt from the sender code:
```cpp
// the issue persists when I change data format or frequency
lsl::stream_info info("SimpleStream", "EEG", 90, 1000.0f, lsl::channel_format_t::cf_double64, "DUMMY_ID");
lsl::stream_outlet outlet(info);
outlet.wait_for_consumers(600);
```

When I print the XML-info for the outlet on the sending machine, everything seems to be correct:
```xml
<?xml version="1.0"?>
<info>
    <name>SimpleStream</name>
    <type>EEG</type>
    <channel_count>90</channel_count>
    <nominal_srate>1000</nominal_srate>
    <channel_format>double64</channel_format>
    <source_id>DUMMY_ID</source_id>
    <version>1,1000000000000001</version>
    <created_at>340,01528632399999</created_at>
    <uid>c7a69651-42af-4749-ae12-706dbae31731</uid>
    <session_id>default</session_id>
    <hostname>schrottpad</hostname>
    <v4address></v4address>
    <v4data_port>16572</v4data_port>
    <v4service_port>16572</v4service_port>
    <v6address></v6address>
    <v6data_port>16573</v6data_port>
    <v6service_port>16573</v6service_port>
    <desc />
</info>
```

The XML for the detected stream_info on the receiving machine is completely useless:
```xml
<?xml version="1.0"?>
<info>
    <name></name>
    <type></type>
    <channel_count>0</channel_count>
    <nominal_srate>0</nominal_srate>
    <channel_format>undefined</channel_format>
    <source_id></source_id>
[...]

PR #25, opened 2019-05-10
Make floating point conversions locale-independent

tstenner:
Fixes at least https://github.com/sccn/liblsl/issues/24 and https://github.com/sccn/lsl_archived/issues/119.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -38,6 +38,8 @@ set (sources
 	src/api_config.h
 	src/cancellable_streambuf.h
 	src/cancellation.h
+	src/cast.h
+	src/cast.cpp
 	src/common.cpp
 	src/common.h
 	src/consumer_queue.cpp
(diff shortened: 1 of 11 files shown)


tstenner:
Paging @sheinke and @maltesen

sheinke:
Hi together, thx so much for fixing this :) I owe you guys, would have never suspected locales. Keep up the good work !

tstenner:
> to_string(float) and to_string(double) should remain separate, so that the precision can be set correctly: 8 for float, and 16 for double. You should also use showpoint: http://www.cplusplus.com/reference/ios/showpoint/. This gives output in significant figures (left and right side of the decimal) rather than precision (right side of the decimal).

Done.

> template <> double from_string should remain separate from template <> float from_string for a slight performance improvement (less casting).

In a loop the version with casts was about 2% slower. The only case where LSL calls `from_string<float>` with multiple values is in `push_chunk_str` with float streams and `pull_chunk_f` with string streams, both of which will so slow enough already that it doesn't matter.

PR #27, opened 2019-06-06
Build CMake object libs for liblsl-static

tstenner:
One of the main obstacles with static libraries is that each source file has to be compiled twice.
[CMake OBJECT libraries](https://cmake.org/cmake/help/v3.5/command/add_library.html#object-libraries) circumvent this by building the objects and using those as sources for the libraries.
In CMake 3.12 the usage was made a lot easier (compiler and linker flags are propagated automatically) and since all CI systems (even the ancient OS X systems on Travis) ship at least 3.12 it's a step we should take.

The Mac tests fail for a reason unknown to me on Azure Pipelines but pass on Travis and our local Macs, so I blame Microsoft for the moment.


cboulay:
This one seems mergeable. Can you please rebase and fix conflict?
While you're at it, maybe try adding the following somewhere sensible if you agree (just so I don't have to make a PR for a simple change):
```
if(${LSL_UNITTESTS} AND NOT ${LSL_BUILD_STATIC})
    set(LSL_BUILD_STATIC ON)
    message(WARNING "Enabling static lib builds as it is required by unit tests.")
endif()
```

tstenner:
I did something a bit different: the static library is built if either LSL_BUILD_STATIC or LSL_UNITTESTS is set, but it's only exported / installed if LSL_BUILD_STATIC is set.

PR #30, opened 2019-08-01
Add  lsl-static to installed targets if LSL_BUILD_STATIC option is ON

arthurbiancarelli:
I'm not sure if it makes sense to "install" a static library, but in my project I find useful to do:

```
find_package(LSL)
...
target_link_library(${TARGET_NAME} LSL::lsl-static
```

First time contributing, I hope I fulfill the requirements

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -185,7 +185,12 @@ endif()
 add_executable(lslver testing/lslver.c)
 target_link_libraries(lslver PRIVATE lsl)
 
-install(TARGETS lsl lslver
+set(EXPORTED_TARGETS lsl lslver)
+if(${LSL_BUILD_STATIC})
+	list(APPEND EXPORTED_TARGETS lslboost ${target}-static)
+endif()
+
+install(TARGETS ${EXPORTED_TARGETS}
 	COMPONENT liblsl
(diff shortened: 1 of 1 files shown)


tstenner:
Something like this has been possible before, but we're in the process of overhauling the build regarding static and shared libraries (see #27).

In short: currently, CMake builds each file twice, once for the static and once for the shared library. With the PR, CMake would build a bunch of object files and link them into a static and a shared library, effectively halving the build time.

The PR in itself is fine, but I'd rather address this after #27. What's your use case for this?

arthurbiancarelli:
Ok thanks for your reply it can wait.  It would be a nice add-on to build object files first.

I'm developing a personal project using LSL library and I wish to link the library statically to have everything bundled in the same executable (and not have dynamic link dependencies which it's causing me some problems, especially on Windows) . Since I'm using CMake for my project too I find it useful to use `find_package` on LSL library.  Hence the need to add the `lsl-static` to the list of exported targets.

cboulay:
@arthurbiancarelli , #27 has now been merged. Do you mind rebasing your PR?

tstenner:
It's already in there, so I'll just close this.

arthurbiancarelli:
perfect thanks

PR #31, opened 2019-08-02
Send multicast requests on all adapters

tstenner:
There have been some more bug reports with streams no / only one stream being found.

Currently, the "I'm looking for a stream"-multicast packets are only sent on one network adapter (on Linux it can be configured otherwise manually, but it ought to work out of the box).

So, this PR adds another config option for all local adapters multicast packets should be sent from, in case it's empty it's generated automatically (currently not yet working on Windows and not tested locally on macOS, but the previously failing CI is happy) and sends the multicast packets once for each configured adapter.

### DIFF
--- a/.github/workflows/cppcmake.yml
+++ b/.github/workflows/cppcmake.yml
@@ -87,6 +87,7 @@ jobs:
               dpkg -I package/liblsl*.deb
            fi
            cmake -E remove_directory package/_CPack_Packages
+           cp testing/lslcfgs/default.cfg .
     - name: upload install dir
       uses: actions/upload-artifact@master
       with:
--- a/src/api_config.cpp
+++ b/src/api_config.cpp
@@ -149,32 +149,29 @@ void api_config::load_from_file(const std::string &filename) {
 		else
 			throw std::runtime_error("This ResolveScope setting is unsupported.");
 
-		multicast_addresses_.insert(
-			multicast_addresses_.end(), machine_group.begin(), machine_group.end());
+		std::vector<std::string> mcasttmp;
+
+		mcasttmp.insert(mcasttmp.end(), machine_group.begin(), machine_group.end());
 		multicast_ttl_ = 0;
 
 		if (scope >= link) {
-			multicast_addresses_.insert(
-				multicast_addresses_.end(), link_group.begin(), link_group.end());
-			multicast_addresses_.push_back("FF02:" + ipv6_multicast_group);
+			mcasttmp.insert(mcasttmp.end(), link_group.begin(), link_group.end());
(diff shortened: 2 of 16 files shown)


jachane99:
Hi,
I'm having this issue for a script in my lab. I'm wondering what's the progress on this issue right now.

tstenner:
Hi @jachane99, this is a somewhat more invasive PR and so it needs more testing than the usual smaller changes. I'd also prefer to have the logging PR (#34) merged before this to make the debug output more useful, but it requires a somewhat newer compiler (i.e. MSVC 2015) so it's unlikely to happen.

If you could replace your liblsl binary with one of the built binaries from this PR ([macOS](https://github.com/sccn/liblsl/suites/403746838/artifacts/1157297), [Ubuntu](https://github.com/sccn/liblsl/suites/403746838/artifacts/1157298), [Windows](https://github.com/sccn/liblsl/suites/403746838/artifacts/1157299)) and see if it works for you, that'd be great.

cboulay:
My only comment/suggestions is in the above review about where to combine the config-provided interfaces and the hardware-polled interfaces.

As far as I can tell, the reset of the code changes are related to storing addresses as `ip`s instead of strings, and pre-emptive error handling. All good stuff.

behinger:
hey! Trying to include this pull request in my linux installation (one direct LAN connection, one university LAN + wifi Connection, should look out for streams on the direct LAN and the wifi) - but so far no success.

1. I checked out the pull request, and ran `standalone_compilation_linux.sh`.
2. I copied the generated `liblsl.so` to pylsl/lib/ and installed it using `pip install .`
3. I tried recording a stream using `examples/ReceiveData.py` -> but I cannot find anything.

I generally don't really know how to debug this issue, any pointers would be greatly appreciated!

cboulay:
@behinger 

> I copied the generated liblsl.so to pylsl/lib/ and installed it using pip install .

Depending on the pylsl source version you're using, this might not do anything. Can you try putting your new `liblsl.so` somewhere it can be found AFTER you do pip install?

See the [pylsl find_liblsl_libraries code](https://github.com/labstreaminglayer/liblsl-Python/blob/master/pylsl/pylsl.py#L1198). For your initial tests, the easiest is to set an environment variable `PYLSL_LIB=path/to/custom/liblsl.so python myscript.py`.

behinger:
Thank you! I did something stupid 🙈 I checked out the PR but didnt switch to the branch. Now it works flawlessly! Super cool!

behinger:
is this PR actually pulled into the release? It appears in the releasenotes, but it didnt work for me. It is also not marked as pulled in this PR.

tstenner:
No idea why it wasn't merged properly, but I have merged it manually (again).

PR #41, opened 2019-10-09
Re-add examples

tstenner:
Discussion is in https://github.com/sccn/labstreaminglayer/issues/37


cboulay:
I think it's now safe to delete https://github.com/labstreaminglayer/App-Examples , correct?

tstenner:
Sure, I have it on at least four other PCs and could re-add it with the whole commit history.

cboulay:
Done. Thanks.

PR #66, opened 2020-03-20
fix compilation with LSL_DEBUGLOG

jfrey-xx:
Hello,

I had to tune `CMakeLists.txt` in order to get liblsl compile when `LSL_DEBUGLOG` is enabled; otherwise I would get the following error:

```
[ 48%] Linking C executable lslver
liblsl.so.1.14.0: undefined reference to `dladdr'
collect2: error: ld returned 1 exit status
CMakeFiles/lslver.dir/build.make:84: recipe for target 'lslver' failed
make[2]: *** [lslver] Error 1
```

Indeed `liblsl/src/loguru/loguru.cpp` includes °<dlfcn.h>`.

Tested on Ubuntu 16.04, with cmake 3.12.0 and gcc 5.5.0.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -165,7 +165,12 @@ target_include_directories(lslboost PUBLIC
 	$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/lslboost>)
 
 
-target_link_libraries(lslobj PRIVATE lslboost)
+# one supplementary dependency for loguru
+if (LSL_DEBUGLOG)
+	target_link_libraries(lslobj PUBLIC ${CMAKE_DL_LIBS}  PRIVATE lslboost)
+else ()
+	target_link_libraries(lslobj PRIVATE lslboost)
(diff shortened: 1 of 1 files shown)


tstenner:
`dl` only has to be linked on Linux, so I added it in 24655a3a2190d54399be5e3be5669127580b9261. Could you confirm that this works for you?

jfrey-xx:
Actually no, with current master compilation still fails,  it seems that `PUBLIC` keyword must be used with `dl`, hence:

```
target_link_libraries(lslobj
	PRIVATE
		lslboost
        PUBLIC 
                $<$<AND:$<BOOL:${LSL_DEBUGLOG}>,$<PLATFORM_ID:Linux>>:dl>
	)
```

works for me :)

tstenner:
That's weird, on both testing machines (Ubuntu 18.04 and 19.10) a private dependency is enough. If added it to the `PUBLIC` section so now it should work just fine.

PR #69, opened 2020-03-30
Lambda initializers

tstenner:
Our oldest compiler (RHEL 5, GCC 4.7 for manylinux1) only has full support for C++11, but partial support for lambda initializers (e.g. `[shared_this = shared_from_this()]() { shared_this->do_cancel(); }`) so we should use it. Also avoid implicit `this` in closures when we have a `shared_ptr` pointing to the object.


tstenner:
Note: the CI for my repository also runs dockerized manylinux1 and manylinux2010 builds and both are succeeding.

PR #74, opened 2020-04-16
Add lsl-static to exported targets if LSL_BUILD_STATIC option is ON

arthurbiancarelli:
Hello. I could not reopen #30. It's been a long time but I just noticed it's missing a small change to export the static library.


tstenner:
I think it should be `OUTPUT_NAME "liblsl${lslplatform}-static"`. I've changed it and merged it.

Issue #83, opened 2020-09-15
StreamInfo ctor does not always return a handle on Win 64 with C#


koehil:
Many days I have been looking at a strange problem, of which I suspect it has to do something with threading. The setup is this: lsllib 1.13 64 bit, Windows 10, source code in C# on .NET3.5. with Visual Studio 2008
In our Windows Forms application, an LSL outlet is to be created on a background thread with Single Threaded Apartment (STA). We use this method (simplified for the purpose of this description):

``` csharp
private void DoConnect()
{
	// we keep a global reference to the outlet object
	if (mOutlet != null)
	{
		// we have added the IDisposable interface to the C# API
		mOutlet.Dispose();
		mOutlet = null;
	}
	
	// TEST MARK 1
	
	// Note: we have renamed StreamInfo to LSLStreamInfo in the LSL API
	liblsl.LSLStreamInfo outStreamInfo = new liblsl.LSLStreamInfo(
		mUniqueStreamName,
		mStreamType,
		mChannelCount,
		mSampleRate,
		liblsl.channel_format_t.cf_double64,
		mSourceID);
	
	// TEST MARK 2
	
	// Sometimes creating the outlet fails. It appeared, that 'outStreamInfo' did
	// not have a valid Handle. Therefore test if the object handle is valid.
	if (outStreamInfo.Handle != IntPtr.Zero)
	{
		// internal helper to construct the XML header
		LSLHelpers.SignalInfoToXMLElement(outStreamInfo.desc(), mMyChannels);
		Trace.WriteLine("LSL: create new outlet:\n" + outStreamInfo.as_xml());
		mOutlet = new liblsl.StreamOutlet(outStreamInfo);

		// Success!				
	}
	else
	{
		Trace.WriteLine("LSL Outlet: PROBLEM: Handle StreamInfo null");
		
		// Fail... :-(
	}
}
```

The first call to DoConnect() for the lifetime of the application runs okay. On some computers, the second call fails. The third time always fails. What happens, is that the LSL StreamInfo constructor returns without handle. In the LSL framework, an exception must have occurred, so that no object handle is returned. Strange enough, looking at the LSL source code, the StreamInfo ctor parameters should not lead to an exception.

But why is the object handle null? Why does it work once or twice, and after that no StreamInfo can be created anymore...?

I have tried a number of things. Firstly, I have put above code and references in a simple console app. From the main thread, but also from a background thread I try to create multiple outlets. No problem. That works!

[...]

PR #84, opened 2020-09-28
Ensure enums are 4 bytes wide as implicitly assumed in previous versions

tstenner:
The original issue is in #83. Basically, the API has several enums that were passed as 4 byte wide parameters to functions. The C# bindings assumed the enums were single bytes, but the 3 bytes padding before the next parameter made it (mostly) work on little endian architectures.

This PR prevents the compiler from reserving a single byte for the enums and aborts compilation with a helpful error message (on most modern compilers) or at least an error with a line number and explanation on older compilers.

### DIFF
--- a/include/lsl/common.h
+++ b/include/lsl/common.h
@@ -1,6 +1,6 @@
 #pragma once
 
-/// @file common.h Global constants for liblsl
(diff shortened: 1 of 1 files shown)


chkothe:
Looks good to merge to me

PR #89, opened 2020-10-30
Concurrency fix and efficiency improvement at low srate from jfrey-xx

chkothe:
This is a merger of #70 along with the queue fix from #71 (and some tweaks).

### DIFF
--- a/src/consumer_queue.cpp
+++ b/src/consumer_queue.cpp
@@ -3,7 +3,6 @@
 #include "sample.h"
 #include "send_buffer.h"
 #include <chrono>
-#include <thread>
 
 using namespace lsl;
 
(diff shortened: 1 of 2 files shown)


tstenner:
So far, looking good. The locking seems fine to me and the benchmark that exhibited the crash / infinite loop runs through just fine.
Two things: there are a bunch of indentation changes that make it hard to see what changed (or who changed a line last with `git blame`) and the commit authorship didn't survive the rebasing.

I've split the PR into two commits (@jfrey-xx's changes and yours) and once the CI is happy I'll merge it.

jfrey-xx:
Thanks a lot for bringing in the changes! It will for sure help a lot here :) Very minor remark, but since I noticed it: in the changelog from https://github.com/tstenner/liblsl/commit/340d517d746cebe480a6072777dcc3398e109fb0 I believe that the race condition is actually from #71 ^^

tstenner:
> Thanks a lot for bringing in the changes! It will for sure help a lot here :) Very minor remark, but since I noticed it: in the changelog from [tstenner@340d517](https://github.com/tstenner/liblsl/commit/340d517d746cebe480a6072777dcc3398e109fb0) I believe that the race condition is actually from #71 ^^

[fixed](https://github.com/tstenner/liblsl/blob/184c3cdbe0686affe9c2e24a168159e4aa5723bf/CHANGELOG)

chkothe:
Ah sorry, my editor was still set to indent with 4 spaces instead of tabs.

PR #98, opened 2020-11-20
Fixes #97 - samples_available now returns samples_available

### LINKED ISSUE #97: samples_available does not return the number of samples, only 1 or 0.
According to the docstring, `samples_available()` is supposed to return the number of samples in the buffer.
https://github.com/sccn/liblsl/blob/1eaaf08ce00fb936c463c029fe91c245560e6891/src/stream_inlet_impl.h#L301-L305

But it returns `(std::size_t)(!data_receiver_.empty());`, and `data_receiver_.empty()` returns a boolean `sample_queue_.empty();`
https://github.com/sccn/liblsl/blob/1eaaf08ce00fb936c463c029fe91c245560e6891/src/data_receiver.h#L71

So samples_available() will only ever return 0 or 1.

cboulay:


chkothe:
PR looks good to me

PR #99, opened 2020-11-27
Replace deprecated Boost / Asio parts

tstenner:
The `cancellable_streambuf` uses deprecated functions in Asio, preventing us from upgrading the included Boost version. This would be especially nice for Unix systems, as the newer Asio releases are faster for single-buffer reads/writes.

`cancellable_streambuf` is based on `basic_socket_streambuf`, the [new implementation](https://github.com/chriskohlhoff/asio/blob/master/asio/include/asio/basic_socket_streambuf.hpp) supports moves but is somewhat scary so I only replaced the deprecated parts.

For future reference: all operating systems *except* macOS quit blocking operations (read, write, connect) and return an error code when the socket is closed from another thread, so subclassing `basic_socket_streambuf` and closing the underlying socket to cancel ongoing operations is nice in theory (and in the length of the implementation), but doesn't work reliably on macOS.


tstenner:
This also fixes an issue in the current implementation where cancelling a read operation returns stale data.

PR #105, opened 2020-12-28
Add C++ API function to retrieve the opaque C API handles

tstenner:
I recently had a situation where I had to push some non-zero terminated strings with known lengths to an outlet with a regular sampling rate, so I only had one timestamp.

The C API had `int32_t lsl_push_chunk_buft(lsl_outlet out, const char **data, const uint32_t *lengths, unsigned long data_elements, double timestamp)`, but it wasn't wrapped in the C++ API and I didn't want to give up the niceties the C++ API offered.

This PR adds a function `raw_lsl_ptr` to both the stream outlet and inlet so the plain C API functions can be called without having to rewrite everything to use the C API.

### DIFF
--- a/include/lsl_cpp.h
+++ b/include/lsl_cpp.h
@@ -766,6 +766,11 @@ public:
 	 */
 	stream_info info() const { return stream_info(lsl_get_info(obj.get())); }
 
+	/// Return a shared pointer to pass to C-API functions that aren't wrapped yet
+	///
+	/// Example: @code lsl_push_chunk_buft(outlet.handle().get(), data, …); @endcode
+	std::shared_ptr<lsl_outlet_struct_> handle() { return obj; }
+
 	/** Destructor.
 	 * The stream will no longer be discoverable after destruction and all paired inlets will stop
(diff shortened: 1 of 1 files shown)


cboulay:
Without testing myself, if a user keeps a reference to the raw handle while the outlet is destructed (which calls `lsl_destroy_outlet(obj)` wherein `delete obj;` is called) then I guess there will be some undesirable behaviour.
I suppose you could add a note in the comments to warn about this, but at the same time I think there's some implied danger when grabbing raw handles.

The PR looks good to me.

tstenner:
> Without testing myself, if a user keeps a reference to the raw handle while the outlet is destructed (which calls `lsl_destroy_outlet(obj)` wherein `delete obj;` is called) then I guess there will be some undesirable behaviour.

Absolutely. The only sane way to fix this is to require C++11 and wrap the pointer in a `shared_ptr`, see #106.

tstenner:
Thanks to @cboulay #106 has been merged and this PR is trivial, so I'm merging this.

PR #113, opened 2021-03-04
Fix API config log verbosity

mesca:
The loguru library was introduced in v1.14.0. This release offers the ability to set the logging verbosity level in the `lsl_api.cfg` file. Unfortunately, the `log.level` setting is currently overriden in `common.cpp`. This PR fixes the issue.


tstenner:
Thanks, that's a real improvement!
I've fixed the formatting, as your editor didn't pick up the configuration files, and added a different message when the default config is getting loaded. The commit is pushed as 741ac631a239612d6bc4f6c322ef55df4b09efe9, so I'll close the PR.

PR #114, opened 2021-03-04
Allow to use system Boost

chausner:
I experimented with updating the code to work with the latest version of Boost (1.75) and also allowing to use the system Boost installation instead of the bundled one. The latter is the recommended approach for package managers like vcpkg and Conan which have their own Boost packages. My intent is to first make the LSL-side more flexible before attempting to create a Conan recipe and improving the existing vcpkg package.

In Boost 1.70 there have been breaking changes in the Asio library which lead to compilation issues in `cancellable_streambuf.h`. I was able to make it compile with some hacks but I haven't tested correctness yet. In any case, the way I solved it is most likely not right and I would love to get some hints from someone who knows the API better.

I saw that the bundled Boost distribution uses a [patch](https://github.com/sccn/liblsl/blob/master/lslboost/boost-thread-windows.patch) on Windows related to timer delays. With an external Boost distribution, this patch would obviously not be included. I am wondering whether it is critical or could maybe even be proposed for inclusion in Boost itself?

Let me know what you think.


tstenner:
> I experimented with updating the code to work with the latest version of Boost (1.75) and also allowing to use the system Boost installation instead of the bundled one. The latter is the recommended approach for package managers like vcpkg and Conan which have their own Boost packages. My intent is to first make the LSL-side more flexible before attempting to create a Conan recipe and improving the existing vcpkg package.

Much appreciated :+1: 

One of the reasons we're bundling boost (or rather, lslboost) is that some programs, most notably Matlab, also link to various versions of Boost and that leads to all sorts of hard to debug issues.
Newer versions of Boost.System are header only and I have replaced most other usages of Boost with the C++11 stdlib.
The only remaining part is #103, but that's still experimental.

> I saw that the bundled Boost distribution uses a patch on Windows related to timer delays. With an external Boost distribution, this patch would obviously not be included. I am wondering whether it is critical or could maybe even be proposed for inclusion in Boost itself?

In the past, Boost has also needed some patches, e.g. so `sleep(.001)` isn't automatically translated to `sleep(.03)`, but as part of the move to C++11 that isn't needed any more.

> In Boost 1.70 there have been breaking changes in the Asio library which lead to compilation issues in cancellable_streambuf.h. I was able to make it compile with some hacks but I haven't tested correctness yet. In any case, the way I solved it is most likely not right and I would love to get some hints from someone who knows the API better.

I'm not a chef on TV, but I've always wanted to say that I've prepared something: #99 has some more fixes than your PR, but it needs some more testing in the real world.

chausner:
Great to hear you are already working on some of the things I tried. I guess I should have checked the existing issues more thoroughly first. :) In this case, I'll wait until the work you've started is merged and will then have a look again at the remaining issues.

tstenner:
I have merged the Asio changes, so if you could rebase your PR I can take a look at it. With my recent changes, the number of accidentally exported Boost symbols is down from several hundreds to a handful from Boost.Archive and Boost.Thread, so in the near future it should be possible to just use a (recent) system install of boost.

chausner:
> I have merged the Asio changes, so if you could rebase your PR I can take a look at it. 

I have rebased onto master.

CMakeLists.txt
```diff
@@ -156,20 +157,49 @@ set_source_files_properties("thirdparty/loguru/loguru.cpp"
...
+if(LSL_BUNDLED_BOOST)
+	message(STATUS "Using bundled Boost")
+		
+	add_library(lslboost OBJECT
+		lslboost/serialization_objects.cpp
+	)
+	target_link_libraries(lslboost PUBLIC Threads::Threads)
+	target_compile_features(lslboost PUBLIC cxx_std_11 cxx_lambda_init_captures)
+
+	target_compile_definitions(lslboost
```
tstenner, line 169: This block should be after the `if(LSL_BUNDLED_BOOST)` as it is mostly identical to the one in the branch below (`PUBLIC` sets flags for the targets `PRIVATE` compilation as well as the `INTERFACE` flags, so it doesn't hurt to reuse this for a target without any source files of its own).

CMakeLists.txt
```diff
@@ -156,20 +157,49 @@ set_source_files_properties("thirdparty/loguru/loguru.cpp"
...
+	)
+	add_library(lslboost INTERFACE)
+	target_compile_definitions(lslboost
+		INTERFACE
+			BOOST_ALL_NO_LIB
+			BOOST_ARCHIVE_SOURCE
+			BOOST_ASIO_STANDALONE
+			#BOOST_ASIO_SEPARATE_COMPILATION
+			BOOST_THREAD_DONT_PROVIDE_INTERRUPTIONS
+			$<$<PLATFORM_ID:Windows>:_WIN32_WINNT=${LSL_WINVER}>
```
tstenner, line 192: This is already handled in the branch at line 206

CMakeLists.txt
```diff
@@ -156,20 +157,49 @@ set_source_files_properties("thirdparty/loguru/loguru.cpp"
...
+			lslboost=boost # allows the LSL code base to work with the original Boost namespace/headers
+	)
+	target_link_libraries(lslboost
```
tstenner, line 195: This is already linked in in line 201

tstenner (changes requested):
Apart from the duplicated fragments I marked this looks good to me.

One other thing: this will lead to all sorts of fun if loaded from an application shipped with a different, private copy of boost (most notably Matlab). Could you add something like `LSLBOOST:${LSL_BUNDLED_BOOST}` to the `LSL_INFO_STR` so the end user can run `lsl_library_info()` to see which boost is linked in?

CMakeLists.txt
chausner: Removed in b9d2b96b57feda4f244f2b6002e8225497a4b337

CMakeLists.txt
chausner: Removed in b9d2b96b57feda4f244f2b6002e8225497a4b337.

CMakeLists.txt
chausner: I moved the block further down as part of b9d2b96b57feda4f244f2b6002e8225497a4b337. Please have a look if this is what you proposed.

chausner: This fails with these errors:
>CMake Error at CMakeLists.txt:191 (target_compile_definitions):
  target_compile_definitions may only set INTERFACE properties on INTERFACE
  targets


>CMake Error at CMakeLists.txt:224 (target_link_libraries):
  INTERFACE library can only be used with the INTERFACE keyword of
  target_link_libraries


>CMake Error at CMakeLists.txt:229 (target_compile_definitions):
  target_compile_definitions may only set INTERFACE properties on INTERFACE
  targets

chausner: I reverted the changes in 512109086acc608acd1851f69d971a681812a8e4 and also fixed some other related issues. This version works for me when using separately installed boost libraries.

tstenner: Ah, too bad. I think there's a more elegant solution for that, but I won't get around to try it this week.
I like the direction this is taking, but I need some more time to see if there's anything left to make it clearer / shorter.

chausner: Thanks, take your time. I guess it would be more concise to do it how it was previously, without splitting up the OS-dependent flags.

tstenner:
After the change to plain Asio was merged, there was only Boost.Serialization left. With the new unit tests, the replacement code for the subset of Boost.Serialization we're actively using is already better tested on x64 than the old code, so I'm confident that it's an alternative to the real Boost.Serialization (especially since your PR builds against new versions by default).
After enabling this by default with system boost, your PR is a bit shorter (https://github.com/tstenner/liblsl/commit/9081e72fa0ab86498bfa56d38534190f030efb97):

``` patch
@@ -27,2 +27,3 @@ option(LSL_OPTIMIZATIONS "Enable some more compiler optimizations" ON)
 option(LSL_UNITTESTS "Build LSL library unit tests" OFF)
+option(LSL_BUNDLED_BOOST "Use the bundled Boost by default" ON)
 option(LSL_BUNDLED_PUGIXML "Use the bundled pugixml by default" ON)
@@ -167,22 +168,30 @@ find_package(Threads REQUIRED)
 # create the lslboost target
-add_library(lslboost OBJECT
-       lslboost/serialization_objects.cpp
-)
-target_link_libraries(lslboost PUBLIC Threads::Threads)
+add_library(lslboost INTERFACE)
+if(LSL_BUNDLED_BOOST)
+       message(STATUS "Using bundled Boost")
+       target_include_directories(lslboost SYSTEM INTERFACE
+               $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/lslboost>)
+else()
+       message(STATUS "Using system Boost")
+       find_package(Boost REQUIRED)
+       target_compile_definitions(lslboost
+               INTERFACE
+                       lslboost=boost # allows the LSL code base to work with the original Boost namespace/headers
+       )
+       target_link_libraries(lslboost INTERFACE Boost::boost Boost::disable_autolinking)
+endif()
 
-if(LSL_SLIMARCHIVE)
+if(LSL_SLIMARCHIVE OR NOT LSL_BUNDLED_BOOST)
        message(STATUS "Using shim instead of full Boost.Archive")
-       target_compile_definitions(lslboost PUBLIC SLIMARCHIVE)
+       target_compile_definitions(lslboost INTERFACE SLIMARCHIVE)
+elseif(LSL_BUNDLED_BOOST)
+       # compile Boost.Serialization objects as part of lslobj
+       target_sources(lslobj PRIVATE lslboost/serialization_objects.cpp)
 endif()
-target_compile_features(lslboost PUBLIC cxx_std_11 cxx_lambda_init_captures)
 
-target_compile_definitions(lslboost
-       PUBLIC
-               BOOST_ALL_NO_LIB
-)
-target_include_directories(lslboost SYSTEM PUBLIC
-       $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/lslboost>)
+target_compile_definitions(lslboost INTERFACE BOOST_ALL_NO_LIB)
 
 # target configuration for the internal lslobj target
-target_link_libraries(lslobj PRIVATE lslboost)
+target_link_libraries(lslobj PRIVATE lslboost Threads::Threads)
+target_compile_features(lslobj PUBLIC cxx_std_11 cxx_lambda_init_captures)
 target_include_directories(lslobj
```

chausner:
> After enabling this by default with system boost, your PR is a bit shorter ([tstenner@9081e72](https://github.com/tstenner/liblsl/commit/9081e72fa0ab86498bfa56d38534190f030efb97)):

Thanks! I updated the PR to match your suggested changes.

tstenner (approved):

tstenner:
Thanks for the PR and your patience :+1:

The CI is happy, but I had to rebase and push the PR by hand (a9a06c78d4218b76e1c906c04452eec697f02f8e).

chausner:
@tstenner 

I have an updated vcpkg port ready and also a Conan recipe:
https://github.com/microsoft/vcpkg/pull/21284
https://github.com/conan-io/conan-center-index/pull/8435

Neither of them has been reviewed yet but I expect them to pass review quickly. As soon as we have a new stable release of liblsl, I can update/publish the ports.

Issue #115, opened 2021-03-10
Don't build as a shared library if LSL_BUILD_STATIC is set


chausner:
Currently, enabling `LSL_BUILD_STATIC` allows to build LSL as a static library but the names of the generated libraries are different (-static suffix) and shared builds are still getting created. The latter is an issue for the vcpkg port because vcpkg poses quite strong restrictions on what should be part of the package and it does not allow shared libraries to be part of it when static builds are enabled. Also, simply deleting the shared library files does not work because there are references to it in the generated CMake target. For this reason, the vcpkg port only supports LSL as a shared library at the moment.

Also for Conan, builds are usually either static or shared and there is no real usecase for building them both at the same time.

My suggestion is to change the behavior of `LSL_BUILD_STATIC` to disable building shared libraries completely (or simply use the pre-defined [BUILD_SHARED_LIBS](https://cmake.org/cmake/help/latest/variable/BUILD_SHARED_LIBS.html) variable in CMake). Then we also don't need the -static suffix and the CMakeList.txt should become simpler overall. It would allow me to support static builds in vcpkg and remove one of the hurdles in creating a Conan package.

tstenner:
I'm somewhat torn on this one. On the one hand, I'm all for getting liblsl in distributions and vcpkg is pretty big on Windows. Building the shared and static library separately enables some optimization opportunities for static libraries, the CMakeLists could use some simplification and so far building liblsl and liblsl-static in one go hasn't been needed so far.

On the other hand, I see liblsl as mostly a shared lib, even for C++ projects. For some vendor provided apps it's only possible to get liblsl bugfixes incorporated by swapping out the liblsl shared library and after my limited success with questions regarding their LSL implementation with some vendors I'm not very optimistic that they'll link their app against another liblsl version. The only way to force that would be to re-license liblsl as LGPL but I'm told that's not likely.
(Still, vendors using vcpkg to get liblsl will be more likely to keep liblsl up to date).

Paging @cboulay and @dmedine for further input.

dmedine:
I for one am completely against the idea of chasing the tail of every new package manager and workflow solution. There is a fork button and if people really need this, that or the other thing, they can press it. There are always going to be hoops to jump through to build a C/C++ project that links to multiple libraries and I don't see why it is our task to jump through _all_ of them rather than our users.

[...]

PR #116, opened 2021-03-22
Don't build as a shared library if LSL_BUILD_STATIC is set

tstenner:
PR for #115

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -7,23 +7,29 @@ project (liblsl
 	)
 
 set(CMAKE_CXX_VISIBILITY_PRESET hidden)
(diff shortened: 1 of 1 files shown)


cboulay:
This is pretty straightforward so I don't feel the need to test more than the CI checks. It's also nice when a commit has net-negative code. I approve.

chausner:
That's great! Thanks for implementing my suggestion. Any particular reason why you opted for a custom `LSL_LIB_TYPE` variable instead of using the existing [`BUILD_SHARED_LIBS`](https://cmake.org/cmake/help/latest/variable/BUILD_SHARED_LIBS.html) which gets used by `add_library` automatically when no type is specified?

Issue #117, opened 2021-03-23
Need ability to reset timestamp postprocessor state


cboulay:
I ran into an issue when using `flush` for a period to ignore data while not allowing the data to accumulate in the queue, then when I resumed using `pull_chunk` the timestamps were incorrect because the postprocessing was trying to smooth them with the pre-flush timestamps.

I think it would be good if using `flush` reset the internal postprocessing state.

More generally, there should be a mechanism to reset the post-processing state. This should be done automatically when `flush` is called, also when `set_postprocessing` is called, and we can expose it via the inlet interface.

Inside the post-processor, the state is reset whenever `smoothing_initialized_` is false (a private variable, false on init): https://github.com/sccn/liblsl/blob/c8cb4dfaba980ffab1b0d18fbc033898a459b361/src/time_postprocessor.cpp#L51

So if we can somehow set `smoothing_initialized_` to false then this should reset the state and fix my problem. We could write a function or a setter to expose this. But first I want to see if there are any existing ways to do this. `smoothing_initialize_` will be set to false if the `query_reset_` callback returns true:
https://github.com/sccn/liblsl/blob/c8cb4dfaba980ffab1b0d18fbc033898a459b361/src/time_postprocessor.cpp#L33-L39

Here is where the inlet initializes the post-processor and passes the callbacks, the 3rd is the `query_reset_` callback.
https://github.com/sccn/liblsl/blob/c8cb4dfaba980ffab1b0d18fbc033898a459b361/src/stream_inlet_impl.h#L45-L47

`time_receiver_.was_reset()` is not too helpful other than telling it to look for `was_reset_`, another private variable:
https://github.com/sccn/liblsl/blob/c8cb4dfaba980ffab1b0d18fbc033898a459b361/src/time_receiver.cpp#L68-L73

One way the `was_reset_` variable gets changed is when `reset_timeoffset_on_recovery`
https://github.com/sccn/liblsl/blob/c8cb4dfaba980ffab1b0d18fbc033898a459b361/src/time_receiver.cpp#L201-L208
This is a bit of a dead end; I don't want to "trick" the inlet into thinking we're recovering from a disconnect. But, we can expose another way to change `was_reset_` and that could be helpful.

Proposed changes:
* `time_postprocessor::set_options` should also do `smoothing_initialized_ = false;`
https://github.com/sccn/liblsl/blob/1eaaf08ce00fb936c463c029fe91c245560e6891/src/time_postprocessor.h#L33
[...]

PR #119, opened 2021-03-27
Reset smoothing parameters after disabling postprocessing

tstenner:
PR for #117.

- Commit 1: add unit tests for postprocessing options
- Commit 2: Move out smoothing parameters and smoothing code into a separate class that can be re-initialized and tested easier
- Commit 3: reset postprocessing options when `set_options()` is called
- Commit 4: adjust samples_seen when flushing samples

There's one major changes: the number of encountered samples is only tracked while `proc_dejitter` is active. Previously, samples were counted and enabling dejittering afterwards re-used this sample counter and therefore assumed a larger number of timestamps had been used to update the dejittering parameters.

### DIFF
--- a/src/stream_inlet_impl.h
+++ b/src/stream_inlet_impl.h
@@ -305,7 +305,11 @@ public:
 	std::size_t samples_available() { return data_receiver_.samples_available(); }
 
 	/// Flush the queue, return the number of dropped samples
-	uint32_t flush() { return data_receiver_.flush(); }
+	uint32_t flush() {
+		int nskipped = data_receiver_.flush();
+		postprocessor_.skip_samples(nskipped);
+		return nskipped;
+	}
 
 	/** Query whether the clock was potentially reset since the last call to was_clock_reset().
 	 *
--- a/src/time_postprocessor.cpp
+++ b/src/time_postprocessor.cpp
@@ -3,17 +3,23 @@
 
 #include <cmath>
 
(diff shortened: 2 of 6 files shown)


tstenner:
I think separating `samples_seen` and `samples_since_t0` will make the estimation more robust with dropped samples, but that'll take some further tinkering.

cboulay:
image

Is this consistent with what you'd expect? The blank periods are periods when I was calling `flush` instead of `pull_chunks`. The y-values are 1/np.diff(timestamps).

gist here: https://gist.github.com/cboulay/006bab4dc971cbd9c7b49e7d32c9cd30

cboulay:
For reference, here's the same thing without any post-processing.
image

cboulay:
Updated: https://gist.github.com/cboulay/006bab4dc971cbd9c7b49e7d32c9cd30

This image was generated when using a conda environment containing a lsl.dll built on the previous version:
image

And this one when using the new lsl.dll
image
Don't worry about the constant offset of 0.6 msec - it's just because I'm using t[0] as my offset, even though t[0] itself is subject to postprocessing with no history so is a poor estimate.

So, in terms of fixing my problem, it looks good.

cboulay:
One more time but only looking at the pre-flush stage to keep in scale for proper comparison of postprocessing functionality before and after the change.

Before change:
image

After change:
image


In addition to the above, I read through the code and unit tests, and I'm satisfied.

Issue #129, opened 2021-06-12
Your deb package for ubuntu 18.04 from release v1.14.0 cannot be installed on ubuntu 18.04


tobiasherzke:
Your deb package from release https://github.com/sccn/liblsl/releases/tag/v1.14.0 for ubuntu 18.04 depends on `libgcc-s1 (>= 3.0)`. A `libgcc-s1` package is not available in the ubuntu package sources for 18.04. This unresolvable dependency is probably introduced into your 18.04 deb package because the github actions runner for 18.04 (you use github actions to create this deb) has some non-standard packages installed, which find their way into the dependency list of debs that you create on this system.

I think you have investigated this problem before in issue #93 and found it (incorrectly) invalid: please note that you accidentally analyzed deb package `liblsl-1.14.0-xenial_amd64.deb` in your closing response. `XENIAL` is `16.04`, not `18.04`.

To enable **our** openMHA users to install liblsl for ubuntu 18.04, we provide our own build which does not have this problem here: https://github.com/HoerTech-gGmbH/liblsl/releases/tag/v1.14.0-htch.

Building an ubuntu 18.04 package on github actions with https://github.com/sccn/liblsl/blob/master/.github/workflows/cppcmake.yml still creates the same unresolvable dependency on `libgcc-s1` for us, meaning Github has not fixed the problem in the last 6 months.

To support our users, we need to build liblsl deb packages for ARM CPUs (in addition to the architectures that you provide installers for) anyway. We do this in encapsulated docker containers, see https://github.com/HoerTech-gGmbH/liblsl/blob/main/.github/workflows/bionic.yml. I have extended this build file to not only build for ARM CPUs, but also build for 64-bit PC (x86_64) in an encapsulated container. Building in a docker container means that we control exactly which packages are installed on the build system.

This results in resolvable dependencies in the deb package for 18.04 (all CPU architectures): Our deb package depends on  `libgcc1 (>= 1:3.0)`, which can be resolved from standard ubuntu package repositories, not on libgcc-s1, which cannot be resolved. You can find the my Dockerfile defining the build environment here if you are interested: https://github.com/HoerTech-gGmbH/docker-buildenv/blob/master/liblsl_x86_64-linux-gcc-7/Dockerfile

[...]

PR #131, opened 2021-06-23
Remove Ubuntu 16.04, migrate Ubuntu 18.04 to docker builds.

tstenner:
Binaries produced by the preinstalled toolchain depend on `libgcc_s1` which isn't in the default repositories (many thanks to @tobiasherzke for the detailed explanation in #129).

This PR adds a config field `docker` that, when set, runs all steps in this container. For Ubuntu 18.04 builds, it selects the `ubuntu-latest` build image and then runs the rest of the build in the `ubuntu:18.04` docker container. This is less flexible than @tobiasherzke's approach, but requires only minimal changes to the CI config and is ~4x faster.

### DIFF
--- a/.github/workflows/cppcmake.yml
+++ b/.github/workflows/cppcmake.yml
@@ -29,14 +29,22 @@ jobs:
       matrix:
         config:
         - {name: "ubuntu-20.04", os: "ubuntu-20.04",   cmake_extra: "-DLSL_BUNDLED_PUGIXML=OFF"}
-        - {name: "ubuntu-18.04", os: "ubuntu-18.04"}
-        - {name: "ubuntu-16.04", os: "ubuntu-16.04"}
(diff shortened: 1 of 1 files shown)


tobiasherzke:
> This PR adds a config field docker that, when set, runs all steps in this container. For Ubuntu 18.04 builds, it selects the ubuntu-latest build image and then runs the rest of the build in the ubuntu:18.04 docker container. This is less flexible than @tobiasherzke's approach, but requires only minimal changes to the CI config and is ~4x faster.

Nice solution! I was not aware of the container config field. I will try to make use of it for my ARM builds, but there I have to prepare the host to be able to execute ARM CPU instructions on the Intel CPU.

I think the speed difference that you see comes from not having to emulate another CPU architecture. I see a similar difference between the ARM builds and the native build in a container. This run of github actions has all 3 in one matrix: https://github.com/HoerTech-gGmbH/liblsl/actions/runs/933023560 The difference there is even 9x to 10x.

PR #132, opened 2021-06-25
Move sample copying/conversion to sample.cpp, use template overloading

tstenner:
A sample has a `char*` array to store the values, but the size and type isn't known at compile time.
The values are iterated over in a cast-happy C++98 for-loop: `for (std::string *p = (std::string *)&data_, *e = p + num_channels_; p < e; ++p)`. Also, sample data can be copied in / out in formats different from the native sample format.

In order to simplify the code, this PR adds a helper class and helper methods, so `for (std::string *p = (std::string *)&data_, *e = p + num_channels_; p < e; ++p) { *p = …; }` becomes `for (auto &val : samplevals<std::string>(*this)) { val = …;}`. It also adds two methods: `conv_from(T* src)` copies the values (converting types if necessary) into the sample data, `conv_into(T* dst)` copies the sample data into `dst` (again, converting the types if necessary). The optimal copying / conversion strategy is found via template overloading.

Even though the implementation is fully in `sample.cpp` instead of `sample.h`, the throughput in a benchmark improved by ~3%.

### DIFF
--- a/src/sample.cpp
+++ b/src/sample.cpp
@@ -2,17 +2,72 @@
 #include "sample.h"
 #include "portable_archive/portable_iarchive.hpp"
 #include "portable_archive/portable_oarchive.hpp"
+#include "util/cast.hpp"
+#include <boost/endian/conversion.hpp>
 
 using namespace lsl;
+using lslboost::endian::endian_reverse_inplace;
 
 #ifdef _MSC_VER
 #pragma warning(suppress : 4291)
 #endif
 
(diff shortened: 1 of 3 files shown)


tstenner:
Some real-world data: with the changes above, the compiler optimized/inlined so much a sampling profiler couldn't detect any of the calls between `stream_outlet_impl::enqueue` and the AVX accelerated memmove instructions.

PR #137, opened 2021-09-15
CMake: remove now unneeded bcrypt from link libraries

tstenner:
As of 70c001868c6ece914e5ec89d7c8690217bddba05, the Windows builds shouldn't depend on `bcrypt` any more.

As far as I know, @chausner @tobiasherzke are most likely to find a niche compiler on Windows that still needs it. It's not urgent at all, but if you could check it, it'd be great.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -198,7 +198,7 @@ if(UNIX AND NOT APPLE)
 	endif()
 elseif(WIN32)
 	target_link_libraries(lslobj PRIVATE iphlpapi winmm)
-	target_link_libraries(lslboost PRIVATE bcrypt mswsock ws2_32)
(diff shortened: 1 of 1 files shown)


tobiasherzke:
Thanks for the consideration. Have built e8b5ffd with MinGW [here](https://github.com/HoerTech-gGmbH/liblsl/tree/nobcrypt) to see if it causes linker problems: Build is green.

Have not tried anything else, we are still using 1.14.0, but will be planning how/when to update to 1.15.2 next week. I may contact you and sccn with suggestions regarding the liblsl `soname` and/or debian packaging depending on how we plan to proceed.

tstenner:
Thanks for testing :+1:

PR #139, opened 2021-09-20
CMake: add object libraries to the export set for static builds

### LINKED ISSUE #138: CMake install broken for static builds
When setting `LSL_BUILD_STATIC=ON`, the CMake install command fails with:
```
CMake Error: install(EXPORT "LSLTargets" ...) includes target "lsl" which requires target "lslobj" that is not in any export set.
CMake Error: install(EXPORT "LSLTargets" ...) includes target "lsl" which requires target "lslboost" that is not in any export set.
CMake Error in CMakeLists.txt:
  export called with target "lsl" which requires target "lslobj" that is not
  in any export set.


CMake Error in CMakeLists.txt:
  export called with target "lsl" which requires target "lslboost" that is
  not in any export set.
```

Seems commit 730e13050142f7140a8b3de85b2a2b164eb1bc0e broke it.

tstenner:
For shared builds, the linker combines all object files and CMake discards the private properties of the `lslboost` and `lslobj` targets.
For shared libraries, it just combines the object files and adds references to the interface properties of the two object libraries, so they need to be imported.

Copying the properties manually is messy and error-prone, so this PR adds the two object libraries to the export this. Hopefully fixes #138.

### DIFF
--- a/.github/workflows/mingw_static.yml
+++ b/.github/workflows/mingw_static.yml
@@ -1,0 +1,55 @@
+name: MinGW Windows static test
+
+on:
+  push:
+    branches: ['*']
+    tags: 
+    paths_ignore: ['docs/**', '.travis.yml']
+  pull_request:
+
+jobs:
+  build:
+    name: MinGW batteries-included
+    runs-on: windows-latest
+    
+    defaults:
+      run:
+        shell: 'msys2 {0}'
+
+    steps:
+    - uses: actions/checkout@v2
+    - uses: msys2/setup-msys2@v2
+      with:
+        release: false
+    - name: Configure CMake
+      run: |
+           cmake --version
+           cmake -S . -B build \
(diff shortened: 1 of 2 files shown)


tstenner:
Paging @chausner

cboulay:
Also need to add `-DLSL_BUILD_STATIC=ON` to CI scripts.

chausner:
Yes, this seems to fix the issue, thanks! 👍

tstenner:
> Also need to add `-DLSL_BUILD_STATIC=ON` to CI scripts.

Well, static builds are not something I would want to encourage, but it needs to be tested. Windows has the biggest issues with the whole `dllexport`/`dllimport` and `.dll` / `.lib` shenanigans, so I've added a separate MinGW build in 
4e7d4c8810ef79dade0ee8ae18cd8a80b38e4526. The resulting packages are huge, but it covers MinGW and static builds at once.

PR #145, opened 2021-09-29
Added socket buffer size options to api_config

cboulay:
This allows setting the TCP socket buffer sizes via lsl_api.cfg

I think the default socket buffer size is different on every platform so instead of assuming a default I just set the default to `0` then only update the socket with the API value if it's larger than 0.


tstenner:
LGTM :+1:

PR #146, opened 2021-09-29
double-typed buff_len

cboulay:
This adds new methods to the C API and an overloaded method to the C++ API to allow the user to set the buffer length with a double instead of an integer. This is necessary in very-high channel count and samplerate situations where even 1-second is too much. Obviously a sub-1-second buffer means that the system won't be able to recover from a disconnect without data loss but at these data rates that isn't really possible anyway (unless we want to start buffering to disk).


tstenner (approved):

cboulay:
@tstenner had a good idea to make this `lsl_create_outlet_ex` and `lsl_create_inlet_ex`, where the new additional argument can receive bitwise-OR'd flags. So in common.h, we'd have
```C++
typedef enum {
    transp_default = 0,                        // keep legacy behavior
    transp_buffsize_samples = 1,       // the supplied int is in samples
    transp_buffsize_thousandths = 2, // Indicates whether the supplied int should be (double)/1000
    transp_synch = 4,                          // the outlet will use synchronous (blocking) zero-copy pushes
    transp_disable_endianness = 8,   // disable endianness conversions
} lsl_transport_options_t;
```

cboulay:
This is such a trivial detail, but do we have any consistency as to whether the common short-form for 'buffer' should be `buf` or `buff`?

tstenner (changes requested):
Few minor things, but other than that an excellent addition

tstenner:
Rebased and merged in 0b35dba3464c22bc707fef88d18483c597519e81

Issue #45, opened 2019-12-16
[macOS] 100% cpu load when running sample app.


CodingMeSwiftly:
I'm using the latest release https://github.com/sccn/liblsl/releases/tag/1.13.0-b13 and include the .dylib in my Xcode project as an embedded framework.

![Screen Shot 2019-12-16 at 20 40 34](https://user-images.githubusercontent.com/6830248/70937338-64d88d80-2044-11ea-9f3b-ccd5c411ecc2.png)

I created a bridging header in oder to be able to use the c library from Swift code.

![Screen Shot 2019-12-16 at 20 42 11](https://user-images.githubusercontent.com/6830248/70937419-8c2f5a80-2044-11ea-938b-442cfbc98e16.png)

I can use the library and invoke lsl_* functions so I assume the lib is linked and included properly in my app target.

However, when I invoke **lsl_resolve_all** the function does not return and the app consumes 100% cpu.

![Screen Shot 2019-12-16 at 21 32 44](https://user-images.githubusercontent.com/6830248/70941071-f5ff3280-204b-11ea-9427-826c62abb052.png)

I understand that the latest release is targeting macOS 10.12. I'm running 10.15(.1). Is that the problem? If so, why are some functions of the lib working just fine but lsl_resolve_all hangs the app? Any idea on how to resolve this?

CodingMeSwiftly:
Curious thing I found while trying to resolve the issue:
If I clone the repo and build the lib on my machine, the resulting .dylib file is 1.6MB.
The .dylib downloaded from the release page linked above is 1.3MB.

Nevertheless, the issue is still there when I try the lib built on my machine.

cboulay:
It's not that surprising that the .dylib's are different sizes. The release one is targeting an older MacOS and is built by a CI, using whatever dependencies the CI pulls in.

FYI, when you create a stream info, that doesn't create the stream, just the metadata. The stream isn't created until you create the outlet. So I don't know if it's what you were expecting, but you wouldn't be able to find "Foo".

Regardless, that shouldn't prevent `lsl_resolve_all` from returning with no streams found after a timeout.

Just so we can isolate the issue, can you try using `lsl_resolve_all` in a simple C/C++ app and see if that works for you? If yes then I really don't know what the issue is. I have zero experience with Swift. If no, then there's some more digging we can do together.

CodingMeSwiftly:
Thank you for the quick response!

[...]

PR #147, opened 2021-09-30
`allow_random_ports`: Let the OS choose an open port

tstenner:
When all configured ports are bound and `allow_random_ports` is enabled, liblsl can try random ports to bind a socket to a port outside the configured range.

This PR changes the socket binding code to open port 0, i.e. let the OS find an open port or abort immediately if there's no free port or no interface to bind to (e.g. as seen in #45).

### DIFF
--- a/src/socket_utils.cpp
+++ b/src/socket_utils.cpp
@@ -23,32 +23,25 @@ uint16_t bind_port_in_range_(Socket &sock, Protocol protocol) {
 		if (!ec) return port;
 	}
 	if (cfg->allow_random_ports()) {
-		for (int k = 0; k < 100; ++k) {
-			uint16_t port = 1025 + rand() % 64000;
-			sock.bind(typename Protocol::endpoint(protocol, port), ec);
-			if (ec == lslboost::system::errc::address_in_use) continue;
-			if (!ec) return port;
-		}
+		// bind to port 0, i.e. let the operating system select a free port
+		sock.bind(typename Protocol::endpoint(protocol, 0), ec);
+		// query and return the port the socket was bound to
+		if (!ec) return sock.local_endpoint().port();
 	}
-	return 0;
+	throw std::runtime_error(
+		"All local ports were found occupied. You may have more open outlets on this machine "
+		"than your PortRange setting allows (see "
+		"https://labstreaminglayer.readthedocs.io/info/network-connectivity.html"
+		") or you have a problem with your network configuration.");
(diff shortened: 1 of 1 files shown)


cboulay:
Can you point to any documentation that states using port 0 will allow the OS to find an open port?

I ran into a dead end from the official docs: https://www.boost.org/doc/libs/1_77_0/doc/html/boost_asio/reference/ip__basic_endpoint/basic_endpoint/overload2.html

tstenner:
> I ran into a dead end from the official docs: https://www.boost.org/doc/libs/1_77_0/doc/html/boost_asio/reference/ip__basic_endpoint/basic_endpoint/overload2.html

It's not in the Asio docs, but the Asio docs leave out a lot of the underlying transports.

For Windows, it's in the documentation for the [`bind()`](https://docs.microsoft.com/en-us/windows/win32/api/winsock/nf-winsock-bind) function:

> For TCP/IP, if the port is specified as zero, the service provider assigns a unique port to the application from the dynamic client port range.

For Linux, it's a bit scattered, but documented in multiple places:

[man ip(7)](https://man7.org/linux/man-pages/man7/ip.7.html):

> ip_local_port_range (since Linux 2.2)
              This file contains two integers that define the default
              local port range allocated to sockets that are not
              explicitly bound to a port number—that is, the range used
              for ephemeral ports.  An ephemeral port is allocated to a
              socket in the following circumstances:
>
>              *  the port number in a socket address is specified as 0
                 when calling bind(2);

[man bind(2)](https://man7.org/linux/man-pages/man2/bind.2.html)

> IP_BIND_ADDRESS_NO_PORT (since Linux 4.2)
              Inform the kernel to not reserve an ephemeral port when
              using bind(2) with a port number of 0.

and

> ERRORS
> EADDRINUSE
              (Internet domain sockets) The port number was specified as
              zero in the socket address structure, but, upon attempting
              to bind to an ephemeral port, it was determined that all
              port numbers in the ephemeral port range are currently in
              use.

And, somewhat hidden, in the [BSD socket API documentation](https://www.freebsd.org/cgi/man.cgi?ip(4)):

> IP_PORTRANGE may be used to set the port range used for selecting a local
     port number on a socket with an unspecified (zero)	port number.  It has
     the following possible values:

cboulay:
I hope Mac works the same way. I don't think it's an often used feature so we'll probably never get any complaints even if it doesn't.

PR #153, opened 2021-10-04
Synchronous Outlet for zero-copying socket writing

cboulay:
This implements a code path that allows for synchronous (blocked) calls to `push_*` by setting an outlet constructor argument `do_async` to `false` or `0`. In this path, `liblsl` performs 0 (zero) copies on the data, compared to the 2 copies performed wen doing asynchronous writing. This can have dramatic speed / CPU usage improvements when using high bandwidth data (18%->5% in one of my tests).

Note that this contains some commits from the `double_buflen` branch. Currently the only relevant commit to examine is c64aeb9.

Some concerns:

* The time spent in `push_*` depends on how many consumers are added. This can lead to unpredictable user experience unless the user application calls `push_*` from a non-interactive thread.
* tcp_server -- I tried to skip launching the `transfer_samples_thread` but this did not work out. So now we have a thread doing nothing. I will try again.
* I only enabled the `do_async` flag on the very new `lsl_create_outlet_ex` C method, and preserved the API of the existing `lsl_create_outlet` method.
* stream_outlet_impl.h now has `#include <asio/buffer.hpp>` for private member `std::vector<asio::const_buffer> sync_buffs_`. A future revision should change this to `std::vector<const_buffer_p>` defined elsewhere.
* I would like to implement an outlet method to `push_buffers_directly(std::vector<buffer_type> buffers, double timestamp, std::vector<std::size_t> frame_indices)` where `frame_indices` gives the indices in `buffers` that start a new frame (frame := multi-channel sample). The goal is to enable the user to push discontiguous frames (e.g., disabled channels still allocated in device buffer) without first copying to a contiguous frame. I'm not sure I actually need this.


cboulay:
Disabling the `transfer_samples_thread` leads to errors on the inlet at chunk boundaries on line https://github.com/sccn/liblsl/blob/cboulay/outlet_sync/src/data_receiver.cpp#L317-L318 . If I leave the do-nothing thread in place then it seems to work fine. :shrug:

cboulay:
Note to self: disable synchronous-write-mode when data format is string or when byte order is not default.

cboulay:
I made some changes based on your suggestions. However, the multiple-client code is probably still broken. I had to deprioritize that due to time constraints. I hope to get back to this soon.

cboulay:
I discovered that [`outlet.have_consumers`](https://github.com/sccn/liblsl/blob/tstenner/outlet_sync/src/stream_outlet_impl.cpp#L170) doesn't work properly in sync mode.

cboulay:
Closing in favour of #170

PR #155, opened 2021-10-06
Replace Boost.Asio with plain Asio

### LINKED ISSUE #143: Investigate replacing boost::asio with regular asio
asio is the boost::asio upstream. http://think-async.com/Asio/AsioAndBoostAsio.html

It should be pretty easy to replace boost::asio this way. I don't think we'll get any advantages beyond further removing dependencies on boost, but I think that's a long term goal anyway.

tstenner:
Closes #143.


cboulay:
It would also be good to get @chausner to look at this.

Issue #142, opened 2021-09-27
MSVC warnings: consumer_queue and stream_inlet_impl not aligned 64


cboulay:
After some updates to the core, it's worth taking some time to look at compiler warnings.

```
C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Tools\MSVC\14.26.28801\include\memory(1588): warning C4316: 'std::_Ref_count_obj2<_Ty>': object allocated on the heap may not be aligned 64
          with
          [
              _Ty=lsl::consumer_queue
          ]
  ..\..\..\src\send_buffer.cpp(12): note: see reference to function template instantiation 'std::shared_ptr<lsl::consumer_queue> std::make_shared<lsl::consumer_queue,int&,std::shared_ptr<lsl::send_buffer>>(int &,std::shared_ptr<lsl::send_buffer> &&)' being compiled
```

```
D:\Tools\Neurophys\labstreaminglayer\LSL\liblsl\src\lsl_c_api_helpers.hpp(83): warning C4316: 'lsl::stream_inlet_impl': object allocated on the heap may not be aligned 64
  ..\..\..\src\lsl_inlet_c.cpp(21): note: see reference to function template instantiation 'Type *create_object_noexcept<lsl::stream_inlet_impl,lsl::stream_info_impl&,int,int32_t&,bool>(lsl::stream_info_impl &,int &&,int32_t &,bool &&) noexcept' being compiled
          with
          [
              Type=lsl::stream_inlet_impl
          ]
```

chausner-audeering:
In some versions of MSVC, it leads to an error instead of warning:
[...]

PR #159, opened 2021-12-08
Rework alignment in consumer_queue

tstenner:
alignas(64) wasn't supported at all for dynamically allocated memory (<C++17), leading to compiler warnings (#142) and undefined behavior according to the Clang UBSan.

Instead, reorder the fields and insert additional padding (if needed) between fields that shouldn't share a cacheline.

### DIFF
--- a/src/consumer_queue.h
+++ b/src/consumer_queue.h
@@ -19,6 +19,18 @@ constexpr int CACHELINE_BYTES = 128;
 constexpr int CACHELINE_BYTES = 64;
 #endif
 
+constexpr int _min(int a, int b) { return a > b ? b : a; }
+
+/// calculate the needed padding to separate two members E1 and E2 by at least CACHELINE_BYTES
+template <typename E1, typename E2, typename... T> constexpr std::size_t padding() {
+	return CACHELINE_BYTES -
+		   _min(sizeof(std::tuple<E1, T..., E2>) - sizeof(E1) - sizeof(E2), CACHELINE_BYTES - 1);
+}
+
+template <typename E1, typename E2, typename... T> struct Padding {
+	const char pad[padding<E1, E2, T...>()]{0};
+};
+
 /**
  * A thread-safe producer/consumer queue of unread samples.
  *
@@ -159,24 +171,35 @@ private:
 		return ++x == wrap_at_ ? 0 : x;
 	}
 
-	/// optional consumer registry
-	send_buffer_p registry_;
+	/// current read position
+	std::atomic<std::size_t> read_idx_{0};
+	/// condition for waiting with timeout
+	std::condition_variable cv_;
 	/// the sample buffer
 	item_t *buffer_;
+
+	/// padding to ensure read_idx_ and write_idx_ don't share a cacheline
+	Padding<std::size_t, std::size_t, std::condition_variable, void *> pad;
+
+	/// current write position
+	std::atomic<std::size_t> write_idx_{0};
 	/// max number of elements in the queue
 	const std::size_t size_;
 	/// threshold at which to wrap read/write indices
 	const std::size_t wrap_at_;
-	// whether we have performed a sync on the data stored by the constructor
-	alignas(CACHELINE_BYTES) std::atomic<bool> done_sync_;
-	/// current write position
-	alignas(CACHELINE_BYTES) std::atomic<std::size_t> write_idx_;
-	/// current read position
-	alignas(CACHELINE_BYTES) std::atomic<std::size_t> read_idx_;
 	/// for use with the condition variable
 	std::mutex mut_;
-	/// condition for waiting with timeout
-	std::condition_variable cv_;
+
+	/// optional consumer registry
+	send_buffer_p registry_;
+
+	/// padding to ensure write_ix_ and done_sync_ don't share a cacheline
+#if UINTPTR_MAX <= 0xFFFFFFFF
+	Padding<std::size_t, bool, std::size_t, std::size_t, std::mutex, send_buffer_p> pad2;
+#endif
+
+	/// whether we have performed a sync on the data stored by the constructor
+	std::atomic<bool> done_sync_{false};
 };
 
 } // namespace lsl
--- a/src/consumer_queue.cpp
+++ b/src/consumer_queue.cpp
@@ -8,16 +8,14 @@
 using namespace lsl;
 
 consumer_queue::consumer_queue(std::size_t size, send_buffer_p registry)
-	: registry_(std::move(registry)), buffer_(new item_t[size]), size_(size),
+	: buffer_(new item_t[size]), size_(size),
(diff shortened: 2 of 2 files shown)


src/consumer_queue.h
```diff
@@ -159,24 +163,35 @@ class consumer_queue {
...
-	/// optional consumer registry
-	send_buffer_p registry_;
+	/// current read position
+	std::atomic<std::size_t> read_idx_{0};
+	/// condition for waiting with timeout
+	std::condition_variable cv_;
 	/// the sample buffer
 	item_t *buffer_;
+
+	/// padding to ensure read_idx_ and write_idx_ don't share a cacheline
```
cboulay, line 173: Can you add a comment explaining where the `2 * sizeof(void *)` comes from?  and why it's `3 *` below?

src/consumer_queue.h
```diff
@@ -159,24 +163,35 @@ class consumer_queue {
...
-	alignas(CACHELINE_BYTES) std::atomic<std::size_t> read_idx_;
 	/// for use with the condition variable
 	std::mutex mut_;
-	/// condition for waiting with timeout
-	std::condition_variable cv_;
+
+	/// optional consumer registry
+	send_buffer_p registry_;
+
+	/// padding to ensure read_idx_ and write_idx_ don't share a cacheline
```
cboulay, line 188: I think we're trying to keep `read_idx_`, `write_idx_`, AND `done_sync_` separated, correct? Otherwise this next padding isn't needed.

cboulay:
I'm pretty new to dealing with alignment issues.

My understanding of the original problem is as follows:
* We need to keep some member variables in `consumer_queue` separated in different cachelines for more efficient access. e.g.., the read index and write index are accessed in different threads.
* Up to now, these variables have been kept separate by using `alignas`
* Variables allocated on the heap, such as `consumer_queue`, don't respect alignment.

This PR gets rid of the `alignas` and replaces it with manual padding. This keeps `read_idx_`, `write_idx_`, AND `done_sync_` in different cachelines (probably?), but it doesn't do any actual alignment. They might be far from a boundary. Or, in the worst case, they might span a boundary and `read_idx_`'s later bits might share a 64-byte block with the early bits of `write_idx_`. Is this actually possible? I admit complete ignorance here.
If it's true that they can span a boundary, then is it likely that this change will be slower than the original?

An alternative might be to create the consumer_queue as a global or on the stack instead of the heap, because globals and stack allocation respects alignment. I think this is impossible. For example, `consumer_queue` is used in `data_receiver`'s `consumer_queue sample_queue_`, `data_receiver` is a member of `stream_inlet_impl`, which is part of the API. We can't dictate to users to never allocate an inlet on the heap!

Another alternative is to override the constructor and destructor to do memory-aligned allocation and deallocation. This seems to be what is suggested most often on SO. @tstenner , what do you think about that option?

tstenner:
> This PR gets rid of the `alignas` and replaces it with manual padding. This keeps `read_idx_`, `write_idx_`, AND `done_sync_` in different cachelines (probably?), but it doesn't do any actual alignment.

Correct. They are still aligned to the types natural alignment (4 or 8 bytes), but other than that there's no further alignment necessary. The `consumer_queue` objects themselves are (most probably) aligned to 16 bytes, so they might have one of four positions in a cache line. CPUs fetch/invalidate the entire cache line, independently of the position of the variable within it, so it doesn't matter where it is in the single cache line.

When the offset between one of the three variables is larger than 64 bytes, they might end up more than one cache line apart, but all in all the `consumer_queue` object fits in less cache lines.

> Or, in the worst case, they might span a boundary and `read_idx_`'s later bits might share a 64-byte block with the early bits of `write_idx_`. Is this actually possible? I admit complete ignorance here.

They are still aligned to 4 / 8 bytes (otherwise we'd need something like `#pragma pack(1)`), so they can't be split across cache line boundaries.

> [..] consumer_queue as a global or on the stack instead of the heap [..]. I think this is impossible.

Almost 100% correct. It's possible to reserver `sizeof(consumer_queue) + CACHELINE_BYTES - 1` bytes, placement-new the object to an aligned offset and store the offset so a custom `operator delete` can delete the correct memory region originally allocated. C++17 also makes it easier, but for VS 2015 it's hard to impossible.



> Another alternative is to override `new` and `delete` to do memory-aligned allocation and deallocation. This seems to be what is [suggested most often on SO](https://stackoverflow.com/a/26526794/1256069).

Without something like `aligned_alloc`, `_mm_malloc`, `posix_memalign` or `_aligned_malloc` it requires a workaround as described above.
The C11 `aligned_alloc` isn't even supported by VS 2019 and `std::align_val_t` is C++17.

The main reason aligned memory is important to some people is to fit an entire object into a cache line or for SIMD operations, but we need neither of those.

cboulay:
> They are still aligned to 4 / 8 bytes (otherwise we'd need something like #pragma pack(1)), so they can't be split across cache line boundaries.

OK that's the key thing I didn't know.

With this method we might use up to (64 - {whatever is in the last cacheline}) bytes more memory using this method than if we did more careful alignment, but I think that's small enough to ignore.

The only question I have left is if `padding` should use `n * sizeof(void *)` or `n * sizeof(size_t)` (or maybe even `sizeof(size_) + sizeof(wrap_at_) + ...` in case the types ever change).
I think it's dangerous to use sizeof(void *) if that is larger than sizeof(size_t) because that might mean that we don't pad enough, right?

src/consumer_queue.h
tstenner: While thinking about this I found an even better way to calculate it:
`template <typename E1, typename E2, typename... T> struct Padding` inserts exactly `n` bytes, so that a struct of `{E1; T...; Padding; E2};` has `E1` and `E2` `CACHELINE_BYTES` apart, even when considering padding within this struct.

`Padding<int,int,  bool, double, bool, double, bool, double>` has 16 bytes whereas  
`Padding<int,int,  bool, bool, bool, double, double, double>` has 32 bytes.

src/consumer_queue.h
tstenner: You're right, the comment was wrong. Fixed in the latest commit.

tstenner:
> With this method we might use up to (64 - {whatever is in the last cacheline}) bytes more memory using this method than if we did more careful alignment, but I think that's small enough to ignore.

With `alignas` the padding would be inserted by the compiler, so it would insert at least 56 bytes between each element.

> I think it's dangerous to use sizeof(void *) if that is larger than sizeof(size_t) because that might mean that we don't pad enough, right?

Dangerous in the sense that it would take up a handful of cycles to synchronize the cache lines.
I have fixed the types in the latest commit and adjusted for padding between the existing members.

cboulay:
OK. All approved by me. Please merge when you are ready.

tstenner:
Let's make that when MSVC is ready…

PR #165, opened 2022-03-06
Build MinGW on Windows server 2019

tobiasherzke:
MinGW build on Github Actions Windows Server 2022 runner failed because CMake was not found in PATH from MSYS2 shell.
CMake is found in Github Actions Windows Server 2019 runner.

For MinGW compilation, exact OS version of the build host is irrelevant.


tstenner:
According to the docs, the windows-latest runners have CMake 3.18, but maybe it's not on the path. For now we can just use the old runners, though.

PR #174, opened 2022-08-16
Load log settings before all other settings.

pmaanen:
Loading the log settings last causes heaps of INFO messages to be
displayed by lsl::get_local_interfaces() even if the user has set log level to -3.
Changing the order causes less ui clutter and respects the users choices.

### DIFF
--- a/src/api_config.cpp
+++ b/src/api_config.cpp
@@ -105,12 +105,6 @@ void api_config::load_from_file(const std::string &filename) {
 		} else
 			loguru::g_stderr_verbosity = log_level;
 
-		// log config filename only after setting the verbosity level
-		if (!filename.empty())
-			LOG_F(INFO, "Configuration loaded from %s", filename.c_str());
-		else
-			LOG_F(INFO, "Loaded default config");
-
 		// read out the [ports] parameters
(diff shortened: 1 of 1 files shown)


src/api_config.cpp
```diff
@@ -92,6 +92,25 @@ void api_config::load_from_file(const std::string &filename) {
...
+
+		std::string log_file = pt.get("log.file", "");
+		if (!log_file.empty()) {
+			loguru::add_file(log_file.c_str(), loguru::Append, log_level);
+			// don't duplicate log to stderr
+			loguru::g_stderr_verbosity = -9;
+		} else
+			loguru::g_stderr_verbosity = log_level;
+
+		// log config filename only after setting the verbosity level
```
cboulay, line 108: This log statement should remain at the end. We should only see "Configuration loaded ..." after we've finished parsing the loaded config.

cboulay:
Thanks for the PR. Don't worry about the ubuntu-18.04 or macOS-latest build failures.

I ask that you address the one comment I left inline then this looks good to merge.

cboulay (approved):

PR #181, opened 2022-12-19
Add pthread_getname_np implementation for NDKs missing it

tstenner:
Not yet tested on a live device, but compiles against the latest NDK.


cboulay:
Seems harmless to non-Android users. No objections to merging here.

cboulay:
Approved.

PR #190, opened 2023-04-10
Upgrade Catch2, use FetchContent

cboulay:


.github/workflows/cppcmake.yml
```diff
@@ -35,8 +35,8 @@ jobs:
         - {name: "ubuntu-20.04", os: "ubuntu-20.04",   cmake_extra: "-DLSL_BUNDLED_PUGIXML=OFF"}
         - {name: "ubuntu-18.04", os: "ubuntu-latest",  docker: "ubuntu:18.04" }
         - {name: "ubuntu-22.04", os: "ubuntu-22.04", cmake_extra: "-DLSL_BUNDLED_PUGIXML=OFF" }
-        - {name: "windows-x64",  os: "windows-2019", cmake_extra: "-T v140,host=x86"}
-        - {name: "windows-32",   os: "windows-2019", cmake_extra: "-T v140,host=x86 -A Win32"}
+        - {name: "windows-x64",  os: "windows-2019", cmake_extra: "-T host=x86"}
```
tstenner, line 38: I generally agree with this, but the [compatibility page](https://labstreaminglayer.readthedocs.io/dev/build_env.html?highlight=visual#recommended-compiler-toolchain) still lists VS 2015 as supported.

As a conservative approach, we could update the docs and change it to v141.
Or we add a VS 2017 runner to the manually run configurations (like static mingw) and let the CI system use the latest toolchain available. I'm not up to date on the latest msvc compatibility issues, but this might mean that end users will need a newer msvcruntime dll and some older Windows versions might get dropped.

testing/catch_main.cpp
```diff
@@ -1,5 +1,5 @@
 #define CATCH_CONFIG_RUNNER
-#include <catch2/catch.hpp>
+#include <catch2/catch_all.hpp>
```
tstenner, line 2: IIRC there's a bunch of specific header files that don't pull in some of the more expensive headers. I might have a branch with these somewhere, but for now these are fine

tstenner (approved):
The MSVC version bump is a big change, the rest is fine.

cboulay:
The ubuntu 18.04 docker image has invalid certs and can't use FetchContent. @tstenner , any recommendations?
Ubuntu 18.04 is past its EOL so we could drop it.

tstenner:
Dropping it solves the problem nicely. With 1.16.2 out we could also upgrade the MSVC baseline and finally use C++14 or even C++17. I'll open a PR to start the discussion.

tstenner:
https://github.com/sccn/labstreaminglayer/pull/112

tstenner:
I have merged the C++17 PR, so after a rebase this could be merged as well.

cboulay:
Catch2 V3 got merged in another PR. This PR is now obsolete.

PR #195, opened 2023-05-15
Fix recovery issues

tstenner:
See https://github.com/sccn/labstreaminglayer/issues/105

### DIFF
--- a/src/inlet_connection.cpp
+++ b/src/inlet_connection.cpp
@@ -14,6 +14,7 @@
 
 using namespace lsl;
 namespace ip = asio::ip;
+
 inlet_connection::inlet_connection(const stream_info_impl &info, bool recover)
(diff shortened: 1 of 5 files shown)


mgrivich:
The new method inlet_connection::set_protocols has an error. It as the parameter const stream_info_impl &info, but refers to host_info_ in the method. Every host_info_ in set_protocols needs to be changed to info. 

Other than that, I tested the pull request and it works.

tstenner:
Thanks, fixed it.

cboulay:
I didn't check that the issue is fixed because Matthew already did that.
Code quality is good and CI workflows are passing. Actually, I haven't seen all the CI workflows succeed in a long time so this is good news.

I'm satisfied. Merge at your leisure.

PR #196, opened 2023-05-27
Use native endianness for "portable" byte order

cboulay:


cboulay:
All the pull_request checks passed. @mgrivich says this fixes his problem. Merging.

mgrivich:
Thanks.

PR #211, opened 2024-04-09
Fix buffer size calculation in resolve_streams

morningf:
Fix buffer size calculation in resolve_streams and resolve_stream functions.
`sizeof(buffer)` returns the byte size of buffer, not the number of elements it contains. This should be modified to `sizeof(buffer) / sizeof(lsl_streaminfo)`.

### DIFF
--- a/include/lsl_cpp.h
+++ b/include/lsl_cpp.h
@@ -822,7 +822,7 @@ private:
  */
 inline std::vector<stream_info> resolve_streams(double wait_time = 1.0) {
 	lsl_streaminfo buffer[1024];
-	int nres = check_error(lsl_resolve_all(buffer, sizeof(buffer), wait_time));
+	int nres = check_error(lsl_resolve_all(buffer, sizeof(buffer) / sizeof(lsl_streaminfo), wait_time));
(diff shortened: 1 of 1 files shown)


cboulay:
Your fix is correct, thanks!

I guess the idea is that it is possible for the resolver to return more streams than there's room for in the buffer. The code was previously saying that the buffer had room for 8192 resolved stream-info-pointers when in reality it only had room for 1024. So if the resolver ever returned more than 1024 streams this would have caused a problem.

I'm happy to merge this in, but I'm curious, did you try to resolve more than 1024 streams?!

morningf:
> Your fix is correct, thanks!
> 
> I guess the idea is that it is possible for the resolver to return more streams than there's room for in the buffer. The code was previously saying that the buffer had room for 8192 resolved stream-info-pointers when in reality it only had room for 1024. So if the resolver ever returned more than 1024 streams this would have caused a problem.
> 
> I'm happy to merge this in, but I'm curious, did you try to resolve more than 1024 streams?!

I just want to encapsulate a TypeScript interface for an electron project, so I want to understand the details of these interfaces as much as possible, and thus I discovered this bug. I also think that not really anyone will resolve so many streams.

PR #214, opened 2024-05-22
:bug: implicit type conversion prevented importing the native library

BRomans:
When importing the native library into Unreal Engine (not the plugin) the implicit conversion from int to bool would crash the build. 
With explicit conversion the build succeeds and the functionality is preserved.

### DIFF
--- a/include/lsl_cpp.h
+++ b/include/lsl_cpp.h
@@ -324,7 +324,8 @@ public:
 
 	/// lsl_stream_info_matches_query
 	bool matches_query(const char *query) const {
-		return lsl_stream_info_matches_query(obj.get(), query);
+		int result = lsl_stream_info_matches_query(obj.get(), query);
+		return result != 0; 
 	}
 
 


tstenner:
The C++ standard mandates that integers get implicitly promoted booleans.

Could you paste the error messages you are getting without this workaround?

BRomans:
I thought so too but here what I get

`[1/6] Compile [x64] CortexSystem.cpp
MyProject\Plugins\lsl\Include\lsl_cpp.h(327): error C4800: Implicit conversion from 'int32_t' to bool. Possible information loss
MyProject\Plugins\lsl\Include\lsl_cpp.h(327): note: consider using explicit cast or comparison to 0 to avoid this warning
C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.38.33130\INCLUDE\stdint.h(20): note: see declaration of 'int32_t'`

tstenner:
Did you or the build environment set "Treat warnings as errors" or the `/WX` flag?

tstenner:
I pushed a shorter workaround that should remove the warning for MSVC.

BRomans:
It was all the presets of the build environment

PR #219, opened 2024-05-25
check sample->format_ before attempting to delete

cboulay:
This is a stupid hack that I just want to see if it lowers the occurrence of the CI runner failures that I can't reproduce locally reliably.


cboulay:
This didn't work, at all.

PR #225, opened 2024-11-14
fix crash on macos on network interface change

WesselVS:
When set_option is called on an unavailable socket, a crash occurs (at least on MacOS). I reproduce this by creating a local outlet, run the GetAllStreams example and unplugging my wire.

### DIFF
--- a/src/resolve_attempt_udp.cpp
+++ b/src/resolve_attempt_udp.cpp
@@ -8,6 +8,7 @@
 #include <asio/ip/address.hpp>
 #include <asio/ip/multicast.hpp>
 #include <exception>
+#include <iostream>
 #include <loguru.hpp>
 #include <sstream>
 
@@ -33,12 +34,15 @@ resolve_attempt_udp::resolve_attempt_udp(asio::io_context &io, const udp &protoc
 			e.what());
 	}
 	unicast_socket_.open(protocol);
+
+
 	try {
 		broadcast_socket_.open(protocol);
 		broadcast_socket_.set_option(asio::socket_base::broadcast(true));
 	} catch (std::exception &e) {
 		LOG_F(WARNING, "Cannot open UDP broadcast socket for resolves: %s", e.what());
 	}
+
 	try {
 		multicast_socket_.open(protocol);
 		multicast_socket_.set_option(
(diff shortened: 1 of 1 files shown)


mediashock:
Hey, which version are you using? do you happen to have the compiled library for osx ARM anywhere?

WesselVS:
> Hey, which version are you using? do you happen to have the compiled library for osx ARM anywhere?

https://github.com/WesselVS/liblsl/releases/tag/fix_release

zeyus:
@WesselVS this crash happens on all platforms I've tried. This includes if I spin up a VM or switch Wifi networks, the app will crash, so this seems invaluable to have

zeyus:
@WesselVS I forgot about your PR but I have added (complementary) functionality in a new PR https://github.com/sccn/liblsl/pull/259

I think there is also a larger discussion to be had around interface management, this is not a concern hopefully in most lab environments, but if you add mobile devices into the mix suddenly it can become a problem.

Some issues can be avoided, e.g. if you are using LSL on a laptop, and it has flaky wifi, you can run the experiment over Ethernet, and only bind LSL to the eth interface.

It could be nice to deal with exceptions / interface changes nicely, either with trying to rebind based on the config (with some graceful timeout / check interval) and/or an exposed rebind API method that could be called when required, as long as the app doesn't crash when a change occurs.

cboulay:
Thanks @WesselVS — your repro (local outlet + GetAllStreams, then unplug the wire) and diagnosis pinned a real bug: `set_option(outbound_interface(...))` throwing out of an asio completion handler, which aborts the resolve wave (oneshot) or terminates the process from the continuous resolver's background thread.

This has been addressed as part of a broader resolver overhaul on the `cboulay/meta_resolve_fixes` branch — commit 99de309a, which credits you as co-author. The fix there covers the same root cause with a few extensions:

- Uses the non-throwing `error_code` overload of `set_option` instead of try/catch, so no exception machinery runs on the send path.
- On failure it logs and carries on, so the unicast + broadcast queries on that pass still go out (rather than skipping the whole round) and multicast falls back to the default interface.
- Wraps the continuous resolver's background `io->run()` in the same catch-and-retry loop the outlet IO threads use, so any other stray handler exception can't take the process down either.
- Adds a regression test (`lsl_test_resolve_bad_interface`) that configures an unusable multicast interface and asserts discovery still succeeds.

Because that branch also consolidated the three resolve sockets into one, this PR no longer applies cleanly against it. Closing in favor of the overhaul — much appreciated!

PR #228, opened 2025-01-20
Add LSL_NO_PACKAGING variable

mscheltienne:
Hello, I'm not sure if this is the cleaner way; but I'd like to statically link `liblsl` in a LabRecorder `.deb` package. My binary correctly includes it, and `ldd` shows the dependency on `liblsl` gone with a working LabRecorder. Yet, if I try to install the `.deb`, I get `Depends: liblsl (= 1.16.5-noble) but it is not installable` because `liblsl` still produces a `.deb` file along mine. 

I'd like to add this argument to be able to set `-DLSL_NO_PACKAGING=ON` in the `cmake` configuration.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -349,6 +349,12 @@ if(LSL_BUILD_EXAMPLES)
 	add_subdirectory(examples)
 endif()
 
+option(
+	LSL_NO_PACKAGING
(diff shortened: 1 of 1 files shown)


mscheltienne:
I found another way by first building `liblsl` statically, uploading the `install` artifact and then building `LabRecorder`.

PR #232, opened 2025-04-10
Possibly fix #218 - next->_next might be unavailble.

### LINKED ISSUE #218: Tests that (almost) always fail
- [ ]  MINGW - internal loguruthreadnames
- [ ]  MacOS - internal reuseport on ipv6 multicast

These don't test LSL code directly so IMO it is OK to omit these tests on the problematic platforms conditionally.

zeyus:
When cur is deleted, next might become invalid. I've run this test locally and it passes, but I also couldn't get it to fail with the previous implementation, but I see it failing in the test outputs.

If this invokes the test runner we can see if the issue persisits.

### DIFF
--- a/src/sample.cpp
+++ b/src/sample.cpp
@@ -466,11 +466,13 @@ sample *factory::pop_freelist() {
 }
 
 factory::~factory() {
-	for (sample *cur = tail_, *next = cur->next_;; cur = next, next = next->next_) {
-		if (cur != sentinel()) delete cur;
-		if (!next) break;
-	}
-	delete[] storage_;
+	sample *cur = tail_;
+    while (cur) {
+        sample *next = cur->next_;  // Save next pointer before potentially deleting cur
+        if (cur != sentinel()) delete cur;
+        cur = next;
+    }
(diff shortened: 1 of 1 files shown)


zeyus:
@cboulay for your consideration.

I don't know exactly where it's going wrong, but for example if i close all the inlets in a loop (without deleting them) then the first works, and each subsequent stream_close results in: 
```
(  17.861s) [R_cf_double64   ]      data_receiver.cpp:344    ERR| Stream transmission broke off (Input stream error.); re-connecting...
```

e.g. for the inlets_10, there'll be that message 9 times.

I don't know if this indicates a problem with the inlets sharing some pointer?

How it is now, the tests pass, and my idea of copying the stream_info is probably not at all related to it. 

I'd be happy to make this into a fully fledged PR but I'd prefer to actually understand the problem rather than guess :)

zeyus:
oh and for the sample destructor, obviously performance is super critical, while `.load` should give some additional safety (if that is even where the problem really is), I guess there's a slight overhead as well.

PR #234, opened 2025-05-19
Allow setting of config file path without requiring environment variables.

zeyus:
On non-desktop platforms, often apps are sandboxed and don't have (read/write) access at all to `$HOME`, `/etc/` or even the current working directory. These paths may not exist.

In addition, environment variables don't necessarily make a lot of sense in the context of a mobile app, and the application data paths might contain GUIDs or other aspects of the path that make it difficult to determine at compile-time.

This PR adds the ability to set a configuration file path for `liblsl` provided that the static member function `lsl_set_config_filename` is called before any other LSL function.

In addition, there's also an option to read config directly from a string stream via `set_api_config_content` which is loaded into the `INI` and discarded.

### DIFF
--- a/include/lsl/common.h
+++ b/include/lsl/common.h
@@ -227,3 +227,12 @@ extern LIBLSL_C_API double lsl_local_clock();
  * no free() method is available (e.g., in some scripting languages).
  */
 extern LIBLSL_C_API void lsl_destroy_string(char *s);
+
+/**
+ * Set the name of the configuration file to be used.
+ *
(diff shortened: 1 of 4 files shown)


zeyus:
@cboulay is there something I should do differently for this PR? Or is it not functionality that you would be interested in adding?

cboulay:
Hi @zeyus 
Sorry, this one got auto-closed when I deleted a branch and I failed to retarget it. I'll try to fix that now.

PR #237, opened 2025-07-14
(APPLE) Build a Universal Framework, and add code signing and notarization

cboulay:
This PR attempts to build the lsl library as a (universal) framework instead of a dylib. The primary advantage of putting everything (library, headers, cmake files) inside a framework vs a folder is that it allows us to add entitlements for multicasting. From what I've read, this will make it easier for consumers of lsl.framework to enable multitasking and I hope this will fix some issues with Unity and Apple Vision Pro.

This one is going to take a while to get working on GHA I'm sure while I workout all the certificate and signing/notarizing steps.


zeyus:
@cboulay let me know if you need any help with this one, I managed to get a pipeline running for code signing and publishing an iOS app via TestFlight. It wasn't fun.

zeyus:
@cboulay Sounds good, I can retarget the other PRs. 

For my experiment and some other things, I have iOS and MacOS targets, so I can definitely try linking against the system lib instead of bundling the built dylib. I will need to pull this branch into my fork and rebase my changes onto it though because I'm relying on the runtime API config functionality. This isn't a problem and any issues from that will be resolved when I retarget that PR in particular

cboulay:
I think this is ready for wider testing and integration of other PRs. Here goes...

PR #238, opened 2025-08-16
cmake: allow to disable install and additional targets

jcelerier:
In case of building liblsl as part of another software project
(in my case https://github.com/ossia/score) the library will install lots of unneeded things)


cboulay:
A couple things...

First, the edit to cmake/LSLCMake.cmake must be reverted. This file gets packaged with LSL for installation and can be used by other applications' build systems to assist with common config and install steps. I don't want these other applications' build systems to now require setting a variable just so they can call a function that is optional anyway.
I'm curious though, why did you identify this as something to change? Your other edits should have prevented this function from ever being called, no?

Second, is there no other way to tell cmake not to run the install targets on dependencies? If you are including the source tree then I get the problem, but if you're using `find_package` semantics then I wouldn't expect it to run any of the liblsl install targets.

jcelerier:
also, are the broken unit tests on windows the fault of this PR?

cboulay:
See #248 instead

PR #245, opened 2025-12-05
Feature/runtime config

zeyus:
Re-do of https://github.com/sccn/liblsl/pull/234 but from a dedicated branch and targeting `cboulay/apple_framework`

> On non-desktop platforms, often apps are sandboxed and don't have (read/write) access at all to $HOME, /etc/ or even the current working directory. These paths may not exist.
> 
> In addition, environment variables don't necessarily make a lot of sense in the context of a mobile app, and the application data paths might contain GUIDs or other aspects of the path that make it difficult to determine at compile-time.
> 
> This PR adds the ability to set a configuration file path for liblsl provided that the static member function lsl_set_config_filename is called before any other LSL function.
> 
> In addition, there's also an option to read config directly from a string stream via set_api_config_content which is loaded into the INI and discarded.


cboulay:
@zeyus , sorry, I deleted the cboulay/apple_framework branch. In my experience, whenever I delete a branch that has been merged, any other PRs that are targeting that deleted branch get updated to target the merged-to branch, i.e. `dev` in this case.

I don't know why that didn't happen here. Maybe because the PR is from outside the org?
Anyway, can you retarget on `dev` please?

ETA: Actually I was able to rebase it; I just had to temporarily recreate my cboulay/apple_framework branch.

zeyus:
@cboulay It should now be updated, I just checked the diff and it looks good, no stray files anymore so there should be no conflicts merging to dev

edit: that said you might want to squash merge because of the rebase adding me to all the commits for some reason

cboulay:
See #258 instead

PR #246, opened 2026-01-06
Fix extra garbage sample transmitted when outlet is destroyed

mscheltienne:
@cboulay This might be the origin of this extra sample as described in https://github.com/labstreaminglayer/pylsl/issues/67 and https://github.com/xdf-modules/pyxdf/pull/150

This PR and the text below is AI-generated / guided, I'm not a C-dev and I don't have the bandwidth to learn enough about the language- but it seems like a plausible explanation. Can you have a look?

### Problem

When an LSL outlet is destroyed while inlets are still connected, an extra sample containing garbage data is transmitted to the inlets. This causes two issues:

1. Extra sample in recorded data: The inlet receives one more sample than was actually pushed by the application
2. Corrupted clock offsets: The extra sample can trigger a clock offset measurement with garbage values, which corrupts timestamp synchronization in downstream tools like `pyxdf`.

### Root cause:

In `tcp_server::end_serving()`, a "ping" sample is pushed to wake up blocked transfer threads:

```
send_buffer_->push_sample(factory_->new_sample(lsl_clock(), true));
```

The problem is that `factory_->new_sample()` creates a real sample object with:
- A valid timestamp (`lsl_clock()`)
- `pushthrough = true` (forces immediate transmission)
- Uninitialized channel data (whatever was in the freelist memory)

The transfer thread checks `if (!samp) continue`; to skip "blank" samples, but this only checks if the pointer is null—the ping sample is a valid object, so it gets serialized and transmitted.

### Fix

Push an empty `sample_p()` instead of creating a real sample:

```
send_buffer_->push_sample(sample_p());
```

This works because:
- The empty `sample_p` (null intrusive_ptr) still triggers the condition variable notification, waking up blocked threads
- The existing check `if (!samp) continue`; correctly skips null samples
- No garbage data is ever created or transmitted

### DIFF
--- a/src/consumer_queue.h
+++ b/src/consumer_queue.h
@@ -157,7 +157,7 @@ private:
 	inline static void copy_or_move(sample_p &dst, const sample_p &src) { dst = src; }
 	inline static void copy_or_move(sample_p &dst, sample_p &&src) { dst = std::move(src); }
 	// helper to either move or drop a value, depending on whether a dst argument is given
-	inline static void move_or_drop(sample_p &src) { src.~sample_p(); }
+	inline static void move_or_drop(sample_p &src) { src.reset(); }
 	inline static void move_or_drop(sample_p &src, sample_p &dst) { dst = std::move(src); }
 
 	/// helper to add a delta to the given index and wrap correctly


JesseLivezey:
It looks like this PR would basically revert the last change to that line.
https://github.com/sccn/liblsl/commit/11b6207e4dcea91482a176eb0c431a8330e2451a

Any idea why it was changed back then?

cboulay:
Let's see if we can summon @mgrivich to ask.

mscheltienne:
As expected:
- CI fails on the newly added tests: https://github.com/sccn/liblsl/actions/runs/20797324904
- CI succeeds with the fix: https://github.com/sccn/liblsl/actions/runs/20796984429

I re-applied the fix. Good for review :)

cboulay:
@mscheltienne , Thank you so much for this!

I am not a C++ expert either, but I did look into it and the analysis and suggested changes seem correct. I will merge this in soon.

While I haven't been bit by the corrupt final sample myself, having one of the unit tests crash most of the time has been such a huge headache for me. I'm very excited for this fix.

PR #249, opened 2026-01-11
Add build-only configs for non-framework dylib on macos, including x86_64

cboulay:
Trying to fix https://github.com/conda-forge/liblsl-feedstock/pull/16#issuecomment-3734827034


larsoner:
I haven't had any issues thus far using `macos-15-intel` plus `macos-latest` on several projects using `cibuildwheel`. Think it's worth trying that first to see if it works? Then figuring out the universal framework stuff can be put off, and if that's far enough in the future maybe it's at a point that `intel` support can just be dropped entirely from conda-forge so we don't have to figure out how to deal with `universal` stuff there...

PR #254, opened 2026-01-17
Leverage std::string overloads in pugixml 1.15

### LINKED ISSUE #247: [REQ] Leverage std::string in pugixml 1.15

cboulay:
Close #247

This is a redo of #253 , originally by @myd7349. The only difference is that it is rebased on `dev` after I updated some GHA secrets. Either that, or the fact that this is coming from my account, should bypass the errors related to secret access.

Original text:

I have fixed the compilation error on Ubuntu.

Specifically, this error occurs when `-DLSL_BUNDLED_PUGIXML=OFF` is specified:

https://github.com/sccn/liblsl/blob/63741a91f6461768c5b4e070f6ef8b62069b556d/.github/workflows/cppcmake.yml#L43

pugixml 1.15 added std::string overloads (in fact, std::string_view) for some APIs, but older versions of pugixml do not support them. When liblsl is built against a non-bundled pugixml prior to 1.15, these issues arise. Therefore, to use the new overloads while remaining compatible with older pugixml versions, it is necessary to check whether the `PUGIXML_HAS_STRING_VIEW` macro is defined.

https://github.com/zeux/pugixml/blob/v1.15/src/pugixml.hpp#L445-L447

### DIFF
--- a/src/stream_info_impl.cpp
+++ b/src/stream_info_impl.cpp
@@ -47,7 +47,11 @@ template <typename T> void append_text_node(xml_node &node, const char *name, co
 }
 
 template <> void append_text_node(xml_node &node, const char *name, const std::string &value) {
+#if defined(PUGIXML_HAS_STRING_VIEW)
+	node.append_child(name).append_child(node_pcdata).set_value(value);
+#else
 	node.append_child(name).append_child(node_pcdata).set_value(value.c_str());
+#endif
 }
 
 void stream_info_impl::write_xml(xml_document &doc) {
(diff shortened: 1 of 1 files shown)


cboulay:
I just wanted to verify that the apple tests pass, which they do. We can return to using #253

PR #260, opened 2026-01-27
Allow UID to be reset via C API

zeyus:
This was floating around in my fork and it is useful for dynamically creating a stream where you define all of the properties but it has not yet been associated with an inlet.


cboulay:
Thanks @zeyus! Picked this up in #280, which keeps your commit and adds a small tidy-up (one-line C wrapper) plus a matching `stream_info::reset_uid()` C++ wrapper for API parity. Verified it builds and exports the symbol. Continuing review there — will close this one once #280 lands.

PR #261, opened 2026-01-28
Fix lsl_get_channel_bytes to return 0 for cft_string as documented

myd7349:
The documentation for `lsl_get_channel_bytes` states that it returns 0 for the string type (`cft_string`), but in practice this is not the case:

https://github.com/sccn/liblsl/blob/cdf917d0936d4fc03c4367de8be40347b449e64e/include/lsl/streaminfo.h#L174-L175

This patch fixes the issue by returning 0 for the string type.

### DIFF
--- a/src/stream_info_impl.cpp
+++ b/src/stream_info_impl.cpp
@@ -243,7 +243,7 @@ bool query_cache::matches_query(const xml_document &doc, const std::string &quer
 }
 
 int stream_info_impl::channel_bytes() const {
-	const int channel_format_sizes[] = {0, sizeof(float), sizeof(double), sizeof(std::string),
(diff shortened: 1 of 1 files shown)


myd7349:
The endianness unit test appears to have failed. I will try to fix this issue.

cboulay:
Thanks @myd7349 — this is correct and worth finishing. Picked it up in #281, which keeps your fix and resolves the `tcpserver` endian test failure you ran into. That failure turned out to be expected: with `channel_bytes()` returning 0 for strings, the server stops volunteering an endian conversion and advertises its own host byte order; the string length-prefixes are byte-swapped on the receiving side instead (the data path uses the separate `format_sizes[]` table, which is unchanged). Updated the test to assert the host byte order accordingly. Full internal + exported test suites pass locally. Will close this once #281 lands.

PR #266, opened 2026-03-02
Feature/fix fullinfo error code

zeyus:
There is an issue with getting a stream `fullinfo`.

Because EC is not initialized with `lsl_no_error` the value of `ec` can be undefined, resulting in checks of `ec` incorrectly reporting that there was an error `ec != 0`.

This just adds the same initialization that exists in other functions that use `ec`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,4 @@
 .idea/
 /cmake-build-*/
 /examples/build*/
+.vscode/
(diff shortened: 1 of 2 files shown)


zeyus:
not sure what happened with the formatting (and teh extra commit that snuck in there!)...I'll fix it later today.


Edit: and done!

.gitignore
```diff
@@ -14,3 +14,4 @@
 .idea/
 /cmake-build-*/
 /examples/build*/
+.vscode/
```
zeyus, line 17: This is just to avoid checking in IDE-related config.
This directory was created locally because I usually have auto-format on, but I have disabled it for liblsl because it made the commits change a bunch of lines in every file I touched with the c++ formatter.

PR #268, opened 2026-04-13
Prefer system-provided library for Catch2 

obbardc:
For distro packages it is prefereable to use system-provided packages. Use the system-provided library if available.


tstenner:
LGTM. While we're at it, we might as well change the FetchContent to a newer Catch2 version.

Issue #88, opened 2020-10-24
TCP-based service discovery


yasithmilinda:
Hello,

I'm attempting to build a browser plugin (WASM) for LSL for my research work, but encountered a few issues that is preventing me from doing so.

1) It appears that LSL uses UDP+multicast for service discovery. Unless I go in a WebRTC route, I'm stuck at implementing this at browser-level. Is there any work to build a TCP-based service discovery into LSL?

2) LSL communicates data through pure sockets, and currently do not support WebSockets. Is there any work on adding WebSocket support into it?

Unless I'm mistaken, it's not possible to use LSL in a browser environment without these features. (For WebSockets, however, we could temporarily use the WebSockets proxy included in Emscripten, but it seems very hacky)

Any ideas / suggestions around this?

tstenner:
1. Not that I know of, starting with TCP's lack of multicast support so you'd need some kind of discovery program that offers a list of local streams via e.g. HTTP.
2. There's an issue for that in the labstreaminglayer repository (https://github.com/sccn/labstreaminglayer/issues/34), but I don't know of anyone actively working on it.

PR #279, opened 2026-06-17
Add lab.ResolveOverTCP: a firewall-robust TCP-probe discovery fallback

cboulay:
## Summary

Adds an opt-in, firewall-robust stream-discovery fallback: `lab.ResolveOverTCP` (default **off**). When enabled, a resolve additionally finds streams by connecting to their **TCP data ports** directly, instead of relying solely on UDP multicast/broadcast/unicast.

This complements (and is independent of) #278: where #278 makes UDP discovery itself firewall-friendly, this PR provides a path that avoids UDP discovery entirely for same-machine and known-host setups.

## Why

UDP discovery replies can be silently dropped by stateful firewalls (and UDP multicast can be blocked on many networks). TCP is a symmetric, connection-oriented protocol that a stateful firewall tracks natively, so probing the data ports directly is robust where UDP discovery isn't. The cost is speed — see below.

## How it works

When `ResolveOverTCP` is enabled, the resolver builds a target list of **loopback (`127.0.0.1`/`::1`) + every `KnownPeer`**, each crossed with the `BasePort..BasePort+PortRange` range. Alongside the normal UDP waves it fires a TCP probe burst (`resolve_attempt_tcp`): for each target it connects, sends `LSL:shortinfo`, reads the reply, and stores the resulting `stream_info` (host taken from the dialed endpoint, ports from the reply XML). Probes run through a small worker pool (cap 16), bounded by a deadline and cancelled the moment enough streams are found.

## Cost (documented loudly)

This is **slow**: one connection attempt per port per host, and closed/filtered remote ports each cost a full connect timeout. A `WARNING` is logged when the option is enabled. Best suited to loopback / a small static `KnownPeers` set, not broad discovery.

## Changes

- **`api_config`** — `lab.ResolveOverTCP` option + getter + enable-time warning.
- **`resolve_attempt_tcp`** (new) — async, worker-pool TCP prober mirroring `resolve_attempt_udp`'s structure and cancellation.
- **`resolver_impl`** — builds the TCP target list; fires a throttled TCP burst per wave, guarded against overlapping bursts via an in-flight check.
- **`tcp_server`** — prerequisite fix: the TCP `LSL:shortinfo` responder now keeps the client session alive through the reply write (it previously captured only the `tcp_server`, risking a destroy-mid-write). This path had no consumer before; this PR is the first.

## Test

`testing/int/resolve_over_tcp.cpp` (own executable, like `runtime_config`, since it sets the config singleton): blinds UDP discovery completely (`ResolveScope=machine` with empty `MachineAddresses`, empty `KnownPeers`) and asserts a local stream is still resolved — so the hit can only have come from the TCP probe.

## Testing

Full suite green locally (macOS universal): the new test passes, and with the feature off (default) `discovery` (live UDP resolution, 730 assertions), `network`, `tcpserver`, and `runtime_config` are unaffected.

## Notes

- Opt-in and default-off: no behavior change unless `ResolveOverTCP = 1` is set.
- liblsl config booleans are `1`/`0` (`from_string<bool>` is `str == "1"`), so the setting is `ResolveOverTCP = 1`. The `lslapicfg` docs (separate repo) should document the option and the slowness warning.
- Independent of #278; both touch the resolver subsystem but the changes don't overlap.

### DIFF
--- a/cmake/SourceFiles.cmake
+++ b/cmake/SourceFiles.cmake
@@ -29,6 +29,8 @@ set(lslsources
         src/portable_archive/portable_oarchive.hpp
         src/resolver_impl.cpp
         src/resolver_impl.h
+        src/resolve_attempt_tcp.cpp
+        src/resolve_attempt_tcp.h
         src/resolve_attempt_udp.cpp
         src/resolve_attempt_udp.h
         src/sample.cpp
--- a/src/api_config.cpp
+++ b/src/api_config.cpp
@@ -295,6 +295,7 @@ void api_config::load(INI &pt) {
 	// read the [lab] settings
 	known_peers_ = parse_set(pt.get("lab.KnownPeers", "{}"));
 	session_id_ = pt.get("lab.SessionID", "default");
+	resolve_over_tcp_ = pt.get("lab.ResolveOverTCP", false);
 
 	// read the [tuning] settings
 	use_protocol_version_ = std::min(
--- a/src/api_config.h
+++ b/src/api_config.h
@@ -176,6 +176,17 @@ public:
 	 */
 	const std::vector<std::string> &known_peers() const { return known_peers_; }
 
+	/**
+	 * @brief Whether to additionally resolve streams by probing TCP data ports directly.
+	 *
+	 * When enabled, a resolve also connects to every port in the BasePort..BasePort+PortRange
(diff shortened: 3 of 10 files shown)


cboulay:
This PR might enable resolution via websockets. See #88

PR #280, opened 2026-06-17
Allow UID to be reset via C API (extends #260)

cboulay:
Extends #260 by @zeyus, which exposes the existing `stream_info_impl::reset_uid()` through the C API. Useful for generating a UID on a locally-constructed `stream_info` that has not yet been associated with an inlet.

This PR keeps @zeyus's original commit and adds a small follow-up commit:

- Collapse the `lsl_reset_uid` C wrapper to the one-line style of its siblings and drop a stray blank line.
- Add a matching `stream_info::reset_uid()` wrapper to `include/lsl_cpp.h` for C++/C API parity.

Verified locally on macOS: configures and builds, and `_lsl_reset_uid` is present in the exported symbols of the shared library.

Closes #260.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/include/lsl/streaminfo.h
+++ b/include/lsl/streaminfo.h
@@ -126,6 +126,14 @@ extern LIBLSL_C_API double lsl_get_created_at(lsl_streaminfo info);
 extern LIBLSL_C_API const char *lsl_get_uid(lsl_streaminfo info);
 
 /**
+ * Reset the UID of the stream info to a new random value.
+ *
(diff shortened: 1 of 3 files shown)


zeyus:
@cboulay Wasn't sure where to mention this to you, but if you're interested, my liblsl.dart related paper finally got published in JOSS https://joss.theoj.org/papers/10.21105/joss.10425 :)

PR #285, opened 2026-06-27
feat: allow setting listen address explicitly

Pandapip1:
This allows for the address to be set explicitly, for applications which would prefer to set their own addresses without using the config file.

WIP; currently in the process of being tested

### DIFF
--- a/include/lsl/outlet.h
+++ b/include/lsl/outlet.h
@@ -40,6 +40,12 @@ extern LIBLSL_C_API lsl_outlet lsl_create_outlet(lsl_streaminfo info, int32_t ch
  */
 extern LIBLSL_C_API lsl_outlet lsl_create_outlet_ex(
 	lsl_streaminfo info, int32_t chunk_size, int32_t max_buffered, lsl_transport_options_t flags);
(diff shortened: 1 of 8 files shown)


cboulay:
Hi @Pandapip1 ,

Believe me, I would love this feature. It would fix a bunch of headaches for me. However, I am extremely reluctant to change the API surface like this, especially when it's something that can be handled with the config file (and possibly requiring #278 and #279). Would runtime config changing then be enough?

liblsl's API changes slowly and that's on purpose. There are manufacturers and government agencies who use liblsl in various capacities at least in-part because it is stable.

I realize that this proposed change does not break the existing API. I'm more concerned about putting out a release with this change then discovering some fundamental problem (e.g., related to security which is an increasing concern) and then we have to break the API to remove this.

I'm happy to leave this PR open for a while (indefinitely) for public comment.

Cheers,
Chad

Pandapip1:
Thanks for the early comment; if this is something that's unlikely to be accepted then I'll be prioritizing working on and testing some other things. I'll close this PR but I'll keep the branch; if anyone wants to fork from it they should free to do so.

PR #286, opened 2026-06-27
feat: add pkg-config support

Pandapip1:
Adds pkg-config discovery.

~~WIP; Untested~~ Tested; works

### DIFF
--- a/cmake/Installation.cmake
+++ b/cmake/Installation.cmake
@@ -83,6 +83,20 @@ install(
     DESTINATION ${LSL_CONFIG_INSTALL_DIR}
 )
 
+# Generate pkg-config file
+if(LSL_UNIXFOLDERS AND NOT LSL_FRAMEWORK)
+    configure_file(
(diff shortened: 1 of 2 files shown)


Pandapip1:
Tested; works

cboulay:
There is a small platform-condition issue: `LSL_FRAMEWORK` defaults to `ON` on Linux too, even though an actual framework is only created under `APPLE AND LSL_FRAMEWORK`. Consequently, `if(LSL_UNIXFOLDERS AND NOT LSL_FRAMEWORK)` suppresses pkg-config generation for the default Linux configuration.

Please change the condition to:

```cmake
if(LSL_UNIXFOLDERS AND NOT (APPLE AND LSL_FRAMEWORK))
```

I reproduced this with CMake configured for a Linux target: configuration succeeded with `LSL_UNIXFOLDERS=ON` and the default `LSL_FRAMEWORK=ON`, but no `lsl.pc` was generated. Reconfiguring with `LSL_FRAMEWORK=OFF` generated it. This was a configuration check on macOS, not a native Linux compile.

PR #288, opened 2026-06-29
fix(treewide): correct a few misused/incorrect types

Pandapip1:
See commit descriptions for more details.

On all the platforms that CI currently builds for, ABI backwards compatibility is maintained with a single exception for a single platform for a set of functions that were broken on that platform anyway, see the description of 4637e6c7b18c567531f3d2fdc111bbd3eeb9f15f. Source backward compatibility is maintained except for specifically C++ code that directly calls the C functions that were converted from `char` to `int8_t` (in bbbd115f96f7de480b89f052dde2f5f609f91835) rather than going through the C++ wrapper, which will cause a type error that is easy to diagnose and fix.

Untested; leaving in draft until tested to my own satisfaction.

### DIFF
--- a/include/lsl/inlet.h
+++ b/include/lsl/inlet.h
@@ -168,7 +168,7 @@ extern LIBLSL_C_API double lsl_pull_sample_d(lsl_inlet in, double *buffer, int32
 extern LIBLSL_C_API double lsl_pull_sample_l(lsl_inlet in, int64_t *buffer, int32_t buffer_elements, double timeout, int32_t *ec);
 extern LIBLSL_C_API double lsl_pull_sample_i(lsl_inlet in, int32_t *buffer, int32_t buffer_elements, double timeout, int32_t *ec);
 extern LIBLSL_C_API double lsl_pull_sample_s(lsl_inlet in, int16_t *buffer, int32_t buffer_elements, double timeout, int32_t *ec);
-extern LIBLSL_C_API double lsl_pull_sample_c(lsl_inlet in, char *buffer, int32_t buffer_elements, double timeout, int32_t *ec);
+extern LIBLSL_C_API double lsl_pull_sample_c(lsl_inlet in, int8_t *buffer, int32_t buffer_elements, double timeout, int32_t *ec);
 extern LIBLSL_C_API double lsl_pull_sample_str(lsl_inlet in, char **buffer, int32_t buffer_elements, double timeout, int32_t *ec);
 ///@}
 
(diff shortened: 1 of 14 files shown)


Pandapip1:
Can confirm this compiles and works; manual testing seems to agree with my backwards compatibility findings.

cboulay:
I checked whether the three commits can be split and independently taken onto current `dev` (`ab5e080b`). They can, but only the third is ready to cherry-pick unchanged:

- **`f517b8419c90927250225dc83a79b53742981b97` — explicit `int32_t` in the C++ wrapper:** cherry-picks cleanly on its own, without either preceding commit. On the supported platforms, `int32_t` is `int`, so this is ABI-neutral. This is a small standalone cleanup.
- **`bbbd115f96f7de480b89f052dde2f5f609f91835` — signed-byte types:** independently useful and preserves the C ABI, but is not ready unchanged. It now conflicts in `include/lsl/inlet.h`, and its internal `char` -> `int8_t` overload changes leave the optional legacy C++ wrapper forwarding `char*`/`vector<char>` to incompatible methods. The legacy build shows four new no-matching-overload errors in `legacy_abi.cpp`. Please adapt those forwarding methods while preserving the exported legacy signatures. Changing the C declarations also requires source migration for C++ callers using the C API directly, even though the binary interface is preserved. This should be a separate focused PR.
- **`085f71ed6387053d6a13a2244032c7c22a00616e` — `unsigned long` -> `size_t`:** please leave this out of an ABI-preserving fix. On 64-bit Windows (MSVC and MinGW), `unsigned long` is 32 bits while `size_t` is 64 bits. This is not specific to 32-bit MinGW; on ordinary 32-bit Windows both are 32 bits. In particular, an old x64 caller supplies the fifth `lsl_pull_chunk_*` argument as a 32-bit stack value, while the changed callee reads 64 bits. I confirmed the width difference with Windows-target compilation and the old caller's four-byte store in generated assembly. Keep existing entry-point signatures, or introduce separately named wider APIs with an explicit compatibility strategy.

The full PR's ordinary macOS data-transfer tests passed (1,271 assertions across 15 cases), but those tests do not validate Windows ABI compatibility. My recommendation is to take the third commit independently, revise the first in a focused change, and defer/redesign the second.

cboulay:
I merged #297 which takes 2 of your changes and left the third here because it is ABI breaking and I'll defer that to a major release.

Pandapip1:
Sounds good, does there exist a timeline for when a v2 release might happen?

PR #291, opened 2026-08-18
Fix have_consumers staying true after an idle consumer disconnects (#267)

### LINKED ISSUE #267: lsl_have_consumers always returns 'true'
# Problem
Function `lsl_have_consumers( )` from `lsl_c.h` always returns _true_. This persists even when there are no longer any consumers.

## Environment
Linux display-desktop 6.8.0-101-generic #101~22.04.1-Ubuntu SMP PREEMPT_DYNAMIC Wed Feb 11 13:19:54 UTC  x86_64 x86_64 x86_64 GNU/Linux

gcc (Ubuntu 11.4.0-1ubuntu1~22.04.3) 11.4.0

cmake version 4.2.3

## Test program
[CMakeLists.txt](https://github.com/user-attachments/files/25778027/CMakeLists.txt)
[main.c](https://github.com/user-attachments/files/25778026/main.c)

## Steps to replicate

1. Build test program.
2. Run test program. This will now wait for a consumer.
3. Run `LabRecorder`.
4. From `LabRecorder`:
    a. Select the Test stream.
    b. Start recording.
5. The test program will now begin printing the return value that it receives from repeated calls to `lsl_have_consumers( )`.
6. From `LabRecorder`:
    a. Stop recording.
    b. Deselect the Test stream.
    c. Close the program.
7. Use Ctrl-C to end the test program, which is still running.

### Example session

From terminal with test program,
```
$ ./build/lsl-haveconsumer-test 
2026-03-05 14:42:27.815 (   0.000s) [lsl-haveconsume ]      netinterfaces.cpp:89    INFO| netif 'lo' (status: 1, multicast: 0, broadcast: 0)
2026-03-05 14:42:27.815 (   0.000s) [lsl-haveconsume ]      netinterfaces.cpp:89    INFO| netif 'enp3s0' (status: 1, multicast: 4096, broadcast: 2)
2026-03-05 14:42:27.815 (   0.000s) [lsl-haveconsume ]      netinterfaces.cpp:89    INFO| netif 'enp0s31f6' (status: 1, multicast: 4096, broadcast: 2)
2026-03-05 14:42:27.815 (   0.000s) [lsl-haveconsume ]      netinterfaces.cpp:89    INFO| netif 'lo' (status: 1, multicast: 0, broadcast: 0)
2026-03-05 14:42:27.815 (   0.000s) [lsl-haveconsume ]      netinterfaces.cpp:89    INFO| netif 'enp3s0' (status: 1, multicast: 4096, broadcast: 2)
2026-03-05 14:42:27.815 (   0.000s) [lsl-haveconsume ]      netinterfaces.cpp:102   INFO| 	IPv4 addr: 64010102
2026-03-05 14:42:27.815 (   0.000s) [lsl-haveconsume ]      netinterfaces.cpp:89    INFO| netif 'enp0s31f6' (status: 1, multicast: 4096, broadcast: 2)
2026-03-05 14:42:27.815 (   0.000s) [lsl-haveconsume ]      netinterfaces.cpp:102   INFO| 	IPv4 addr: a23074a
2026-03-05 14:42:27.815 (   0.000s) [lsl-haveconsume ]      netinterfaces.cpp:89    INFO| netif 'lo' (status: 1, multicast: 0, broadcast: 0)
2026-03-05 14:42:27.815 (   0.000s) [lsl-haveconsume ]      netinterfaces.cpp:89    INFO| netif 'enp3s0' (status: 1, multicast: 4096, broadcast: 2)
2026-03-05 14:42:27.815 (   0.000s) [lsl-haveconsume ]      netinterfaces.cpp:105   INFO| 	IPv6 addr: fe80::bab3:e1a5:6cb1:8254%enp3s0
2026-03-05 14:42:27.815 (   0.000s) [lsl-haveconsume ]         api_config.cpp:270   INFO| Loaded default config
LSL version information
    library   1.16
    protocol  1.10
    lib info  git:6ca188c266c21f7228dc67077303fa6abaf2e8be/branch:refs/tags/v1.16.2/build:Release/compiler:GNU-11.3.0/link:SHARED
2026-03-05 14:42:27.815 (   0.000s) [lsl-haveconsume ]             common.cpp:65    INFO| git:6ca188c266c21f7228dc67077303fa6abaf2e8be/branch:refs/tags/v1.16.2/build:Release/compiler:GNU-11.3.0/link:SHARED
lsl_have_consumers 0 return 1
lsl_have_consumers 1 return 1
lsl_have_consumers 2 return 1
lsl_have_consumers 3 return 1
lsl_have_consumers 4 return 1
lsl_have_consumers 5 return 1
lsl_have_consumers 6 return 1
lsl_have_consumers 7 return 1
lsl_have_consumers 8 return 1
lsl_have_consumers 9 return 1
lsl_have_consumers 10 return 1
lsl_have_consumers 11 return 1
lsl_have_consumers 12 return 1
lsl_have_consumers 13 return 1
lsl_have_consumers 14 return 1
lsl_have_consumers 15 return 1
lsl_have_consumers 16 return 1
lsl_have_consumers 17 return 1
lsl_have_consumers 18 return 1
lsl_have_consumers 19 return 1
lsl_have_consumers 20 return 1
lsl_have_consumers 21 return 1
lsl_have_consumers 22 return 1
lsl_have_consumers 23 return 1
lsl_have_consumers 24 return 1
lsl_have_consumers 25 return 1
lsl_have_consumers 26 return 1
lsl_have_consumers 27 return 1
lsl_have_consumers 28 return 1
lsl_have_consumers 29 return 1
lsl_have_consumers 30 return 1
lsl_have_consumers 31 return 1
lsl_have_consumers 32 return 1
^CHandled signal 2
lsl_have_consumers final return 1
```
From terminal with LabRecorder,
```
$ LabRecorder
"LabRecorder.cfg"
"/home/display/Documents/Jira/API-3375/LabRecorder.cfg"
"/home/display/.config/LabRecorder/LabRecorder.cfg"
"/etc/xdg/xdg-ubuntu/LabRecorder/LabRecorder.cfg"
"/etc/xdg/LabRecorder/LabRecorder.cfg"
"/home/display/.local/share/LabRecorder/LabRecorder.cfg"
"/usr/share/ubuntu/LabRecorder/LabRecorder.cfg"
"/usr/share/gnome/LabRecorder/LabRecorder.cfg"
"/usr/local/share/LabRecorder/LabRecorder.cfg"
"/usr/share/LabRecorder/LabRecorder.cfg"
loading config file  "/usr/share/LabRecorder/LabRecorder.cfg"
2026-03-05 14:42:30.315 (   0.055s) [LabRecorder     ]      netinterfaces.cpp:89    INFO| netif 'lo' (status: 1, multicast: 0, broadcast: 0)
2026-03-05 14:42:30.315 (   0.055s) [LabRecorder     ]      netinterfaces.cpp:89    INFO| netif 'enp3s0' (status: 1, multicast: 4096, broadcast: 2)
2026-03-05 14:42:30.315 (   0.055s) [LabRecorder     ]      netinterfaces.cpp:89    INFO| netif 'enp0s31f6' (status: 1, multicast: 4096, broadcast: 2)
2026-03-05 14:42:30.315 (   0.055s) [LabRecorder     ]      netinterfaces.cpp:89    INFO| netif 'lo' (status: 1, multicast: 0, broadcast: 0)
2026-03-05 14:42:30.315 (   0.055s) [LabRecorder     ]      netinterfaces.cpp:89    INFO| netif 'enp3s0' (status: 1, multicast: 4096, broadcast: 2)
2026-03-05 14:42:30.315 (   0.055s) [LabRecorder     ]      netinterfaces.cpp:102   INFO| 	IPv4 addr: 64010102
2026-03-05 14:42:30.315 (   0.055s) [LabRecorder     ]      netinterfaces.cpp:89    INFO| netif 'enp0s31f6' (status: 1, multicast: 4096, broadcast: 2)
2026-03-05 14:42:30.315 (   0.055s) [LabRecorder     ]      netinterfaces.cpp:102   INFO| 	IPv4 addr: a23074a
2026-03-05 14:42:30.315 (   0.055s) [LabRecorder     ]      netinterfaces.cpp:89    INFO| netif 'lo' (status: 1, multicast: 0, broadcast: 0)
2026-03-05 14:42:30.315 (   0.055s) [LabRecorder     ]      netinterfaces.cpp:89    INFO| netif 'enp3s0' (status: 1, multicast: 4096, broadcast: 2)
2026-03-05 14:42:30.315 (   0.055s) [LabRecorder     ]      netinterfaces.cpp:105   INFO| 	IPv6 addr: fe80::bab3:e1a5:6cb1:8254%enp3s0
2026-03-05 14:42:30.315 (   0.055s) [LabRecorder     ]         api_config.cpp:270   INFO| Loaded default config
Missing:  QSet()
2026-03-05 14:42:39.046 (   8.787s) [LabRecorder     ]             common.cpp:65    INFO| git:6ca188c266c21f7228dc67077303fa6abaf2e8be/branch:refs/tags/v1.16.2/build:Release/compiler:GNU-11.3.0/link:SHARED
Opened the stream Test.
Received header for stream Test.
Started data collection for stream Test.
Offsets thread is finished
Wrote footer for stream Test.
Closing the file.
```

daverlon:
Fixes #267

The transfer thread only noticed a disconnected peer after a failed write. Outlets that were not pushing never unregistered the queue. Watch the TCP socket for peer close so an idle outlet unregisters the consumer without waiting for a push.

### DIFF
--- a/src/tcp_server.cpp
+++ b/src/tcp_server.cpp
@@ -16,6 +16,7 @@
 #include <asio/streambuf.hpp>
 #include <asio/write.hpp>
 #include <algorithm>
+#include <atomic>
 #include <condition_variable>
 #include <cstdint>
 #include <cstring>
@@ -261,8 +262,11 @@ private:
 	/// Handler that gets called sending the feedheader has completed.
 	void handle_send_feedheader_outcome(err_t err, std::size_t n);
 
+	/// Detect peer close on the feed socket so idle consumers get unregistered (see #267).
+	void abort_on_peer_close(const std::shared_ptr<consumer_queue> &queue);
+
 	/// Transfers samples from the server's send buffer into the async send queues of IO threads
-	void transfer_samples_thread(std::shared_ptr<client_session> /*keepalive*/,
+	void transfer_samples_thread(std::shared_ptr<client_session> keepalive,
 		std::shared_ptr<consumer_queue> &&queue, int max_samples_per_chunk);
 
 	/// Handler that gets called when a sample transfer has been completed.
@@ -295,6 +299,10 @@ private:
 	int chunk_granularity_{0};
 	/// maximum number of samples buffered
 	int max_buffered_{0};
+	/// set when the consumer closes the feed connection (or sends unexpected data)
+	std::atomic<bool> abort_transfer_{false};
+	/// dummy byte for the peer-close detector's pending read
+	char peer_close_byte_{0};
 
 	// data exchanged between the transfer completion handler and the transfer thread
 	/// whether the current transfer has finished (possibly with an error)
@@ -744,6 +752,7 @@ void client_session::handle_send_feedheader_outcome(err_t err, std::size_t n) {
 
 		// determine transfer parameters
 		auto queue = serv->send_buffer_->new_consumer(max_buffered_);
+		abort_on_peer_close(queue);
 
 		// determine the maximum chunk size
 		int max_samples_per_chunk = std::numeric_limits<int>::max();
@@ -761,13 +770,27 @@ void client_session::handle_send_feedheader_outcome(err_t err, std::size_t n) {
 	}
 }
 
-void client_session::transfer_samples_thread(std::shared_ptr<client_session> /* keepalive */,
+void client_session::abort_on_peer_close(const std::shared_ptr<consumer_queue> &queue) {
+	// The feed is one-way after the handshake; a completed read means EOF or unexpected data.
+	// Either way the consumer is gone. A weak_ptr keeps the queue from staying registered if the
+	// transfer thread has already exited (e.g. after a failed write).
+	std::weak_ptr<consumer_queue> weak_queue = queue;
+	sock_.async_read_some(asio::buffer(&peer_close_byte_, 1),
+		[shared_this = shared_from_this(), weak_queue](err_t err, std::size_t /*n*/) {
+			if (err == asio::error::operation_aborted) return;
+			shared_this->abort_transfer_.store(true, std::memory_order_release);
+			if (auto q = weak_queue.lock()) q->push_sample(sample_p());
+		});
+}
+
+void client_session::transfer_samples_thread(std::shared_ptr<client_session> keepalive,
 	std::shared_ptr<consumer_queue> &&queue, int max_samples_per_chunk) {
 	int samples_in_current_chunk = 0;
-	while (!serv_.expired()) {
+	while (!serv_.expired() && !abort_transfer_.load(std::memory_order_relaxed)) {
 		try {
 			// get next sample from the sample queue (blocking)
 			sample_p samp(queue->pop_sample());
+			if (abort_transfer_.load(std::memory_order_acquire)) break;
 
 			// ignore blank samples (they are basically wakeup notifiers from someone's
 			// end_serving())
@@ -800,6 +823,12 @@ void client_session::transfer_samples_thread(std::shared_ptr<client_session> /* 
 			LOG_F(WARNING, "Unexpected glitch in transfer_samples_thread: %s", e.what());
 		}
 	}
+	// Unregister immediately; cancel the peer-close read so the session can be destroyed.
+	queue.reset();
+	post(*io_, [keepalive]() {
+		asio::error_code ec;
+		keepalive->sock_.cancel(ec);
+	});
 }
 
 void client_session::handle_chunk_transfer_outcome(err_t err, std::size_t len) {
--- a/testing/CMakeLists.txt
+++ b/testing/CMakeLists.txt
@@ -70,6 +70,7 @@ add_executable(lsl_test_exported
 	ext/DataType.cpp
 	ext/discovery.cpp
 	ext/move.cpp
+	ext/outlet.cpp
 	ext/streaminfo.cpp
 	ext/sync_outlet.cpp
 	ext/time.cpp
--- a/testing/ext/outlet.cpp
+++ b/testing/ext/outlet.cpp
@@ -1,0 +1,36 @@
+#include <catch2/catch_all.hpp>
+#include <chrono>
+#include <lsl_cpp.h>
+#include <thread>
+
+// clazy:excludeall=non-pod-global-static
+
+namespace {
+
+bool wait_until_no_consumers(lsl::stream_outlet &outlet, double timeout_sec) {
+	auto deadline =
+		std::chrono::steady_clock::now() + std::chrono::duration<double>(timeout_sec);
+	while (outlet.have_consumers() && std::chrono::steady_clock::now() < deadline)
+		std::this_thread::sleep_for(std::chrono::milliseconds(10));
+	return !outlet.have_consumers();
+}
+
+TEST_CASE("have_consumers becomes false after idle inlet disconnects", "[outlet][basic]") {
+	lsl::stream_info info("have_consumers_idle", "Markers", 1, lsl::IRREGULAR_RATE, lsl::cf_int32,
+		"have_consumers_idle");
+	lsl::stream_outlet outlet(info);
+	auto found = lsl::resolve_stream("name", info.name(), 1, 2.0);
+	REQUIRE(found.size() == 1);
+
+	{
+		lsl::stream_inlet inlet(found[0]);
+		inlet.open_stream(2);
+		REQUIRE(outlet.wait_for_consumers(2));
+		REQUIRE(outlet.have_consumers());
+		// Intentionally do not push samples; disconnect must still be detected (#267).
+	}
+
+	CHECK(wait_until_no_consumers(outlet, 2.0));
+}
+
+} // namespace


cboulay:
The idle-disconnect test passes locally, but I found a concurrency issue that needs addressing before merge.

In `abort_on_peer_close()`, the Asio callback calls `q->push_sample(sample_p())` directly. `consumer_queue::push_sample()` explicitly requires a single producer. Normal sample pushes are serialized by `send_buffer::consumers_mut_`, but this callback bypasses that mutex, so disconnecting while the outlet is pushing can put two producers in the queue simultaneously. `try_push()` uses a load/store of the write index rather than a CAS, allowing both producers to claim the same slot and race on its sample pointer.

Please use a separate cancellation/wakeup mechanism that does not enqueue from a second producer, or serialize the wakeup with all normal pushes. The current idle-only regression does not cover this; please also test disconnects concurrent with active sample production, preferably under ThreadSanitizer.

Validation: built the PR on macOS and ran the new `[outlet]` test; all 4 assertions passed. The race above is a source-level finding, not a claim that the idle test reproduced it.

cboulay:
Merged current `dev` (including #299 and #300) into this branch in 246d147d. The conflict resolution preserves #300’s queue subscription before the handshake and starts this PR’s disconnect watcher using that existing queue after the header write completes. It also retains the updated synchronous socket handoff. This update resolves integration conflicts; it does not yet implement the recommendations below.

I support fixing idle disconnect detection: a marker outlet should stop reporting a consumer that has already disconnected, even if no samples are pushed afterward. The pending asynchronous read is a reasonable event-driven approach, but I recommend revising the cleanup before merging:

1. **Serialize the wakeup with ordinary sample pushes.** `abort_on_peer_close()` currently calls `q->push_sample(sample_p())` directly from the IO callback. `consumer_queue` requires a single producer, while normal outlet pushes are serialized through `send_buffer::consumers_mut_`. This callback bypasses that lock and can race with a real sample push. A small targeted wakeup helper on `send_buffer`, using the same mutex and retaining the queue safely for the operation, would preserve the existing empty-sample wakeup convention without adding a concurrent producer. A dedicated queue cancellation operation is another option, but would be a broader change.
2. **Account for synchronous outlets.** Their sockets are owned by `sync_write_handler` and bypass this watcher. They still retain idle disconnected consumers until a write fails. This needs explicit coverage here or a clearly scoped follow-up with appropriate socket ownership/synchronization.
3. **Preserve #300’s early subscription and watcher cleanup.** Moving subscription back after the handshake would reintroduce the first-sample race. Failed header writes already release the queue; the pending read must also be cancelled when the transfer thread exits.
4. **Expand regression coverage** to explicit close/reopen and disconnecting one of two consumers, plus synchronous idle disconnects when supported. Keep the immediate-first-sample tests from #300.

#299 makes inlet shutdown deterministic, but the outlet still needs to observe the disconnect. Also, this fix removes stale consumers; it cannot guarantee that a consumer satisfying `wait_for_consumers()` is the intended recorder or will remain connected. That stronger experiment-start guarantee needs an application-level readiness exchange.

Validation of this merge: macOS universal build passed; `[outlet],[open],[reopen],[sync]` passed with **1,691 assertions across 17 test cases**. These tests do not establish absence of the producer race above.

daverlon:
Fixed. `send_buffer::wake_consumer()` does the empty-sample push under `consumers_mut_`, and
`abort_on_peer_close()` calls that instead of pushing to the queue directly, so the IO thread
is no longer a second producer.

Before the fix I was able to reproduce the race you described under ThreadSanitizer. It does reproduce once a test pushes concurrently with the disconnect on macOS arm64, AppleClang21, Debug, `-fsanitize=thread`:
```
WARNING: ThreadSanitizer: data race
  Read of size 8 at 0x00011061c828 by thread T1:
    #0  lslboost::intrusive_ptr<lsl::sample>::swap(...)
    #1  lslboost::intrusive_ptr<lsl::sample>::operator=(intrusive_ptr&&)
    #2  lsl::consumer_queue::copy_or_move(..., intrusive_ptr&&)
    #3  lsl::consumer_queue::try_push<intrusive_ptr<lsl::sample>>(...)
    #4  lsl::consumer_queue::push_sample<intrusive_ptr<lsl::sample>>(...)
    #5  lsl::client_session::abort_on_peer_close(...)::$_0::operator()(...)
    [... asio dispatch ...]
    #12 asio::io_context::run()
  Previous write of size 8 at 0x00011061c828 by thread T3 (mutexes: write M0):
    #0  lslboost::intrusive_ptr<lsl::sample>::swap(...)
    #1  lslboost::intrusive_ptr<lsl::sample>::operator=(intrusive_ptr const&)
    #2  lsl::consumer_queue::copy_or_move(..., intrusive_ptr const&)
    #3  lsl::consumer_queue::try_push<intrusive_ptr<lsl::sample> const&>(...)
    #4  lsl::consumer_queue::push_sample<intrusive_ptr<lsl::sample> const&>(...)
    #5  lsl::send_buffer::push_sample(intrusive_ptr<lsl::sample> const&)
    #6  lsl::stream_outlet_impl::enqueue<int>(...)
    #9  lsl::stream_outlet::push_sample(int const*, double, bool)
  Location is heap block of size 576016 allocated by thread T1:
    #1  lsl::consumer_queue::consumer_queue(unsigned long, shared_ptr<send_buffer>)
    #9  lsl::send_buffer::new_consumer(int)
    #10 lsl::client_session::handle_read_feedparams(...)
  Mutex M0 created at:
    #2  std::lock_guard<std::mutex>::lock_guard(std::mutex&)
    #3  lsl::send_buffer::push_sample(intrusive_ptr<lsl::sample> const&)
```

Both threads are in `consumer_queue::try_push` on the same ring buffer slot. TSan identifies
M0 as the lock taken in `send_buffer::push_sample` and annotates only T3 with it; the
`abort_on_peer_close` callback on T1 holds nothing.

To confirm the fix rather than just fail to see the race, I built the pre-fix commit and this
branch with identical flags and ran them interleaved with 550 runs each. Before the change there were 97 queue races in 550 runs, and after it 0 queue races. With the fix all 1100 runs passed, and I wrapped each run in a 120s kill in case the new lock deadlocked and none of them hit it.

Reproduce on 246d147d:
```bash
cmake -S . -B build-tsan -DCMAKE_BUILD_TYPE=Debug -DLSL_UNITTESTS=ON \
  -DCMAKE_C_FLAGS="-fsanitize=thread -fno-omit-frame-pointer" \
  -DCMAKE_CXX_FLAGS="-fsanitize=thread -fno-omit-frame-pointer" \
  -DCMAKE_EXE_LINKER_FLAGS="-fsanitize=thread" \
  -DCMAKE_SHARED_LINKER_FLAGS="-fsanitize=thread"
cmake --build build-tsan --parallel
for i in $(seq 1 500); do
  TSAN_OPTIONS="history_size=7 halt_on_error=0" \
    ./build-tsan/testing/lsl_test_exported \
    "have_consumers becomes false after disconnect during push"
done
```

I initially tried a `consumer_queue::cancel()` but it needs a flag on `consumer_queue` that never resets, and a `data_receiver` uses that class too.

Added three new tests: 
- disconnect during active production
- explicit close/reopen
- disconnecting one of two inlets. 

`[outlet],[open],[reopen],[sync]` is 1710 assertions across
20 cases. The pre-handshake subscription and the `sock_.cancel()` on transfer-thread exit are
untouched.

Synchronous idle disconnects are in #303. I'm not planning to pick that one up, but tell me if you'd rather it land here.

TSan also flags a pre-existing `factory::new_sample()` / `data_receiver::pull_sample_typed()` race, 21 times pre-fix, 17 after. Not from this change, and it's how I know TSan was still live on the fixed build.

cboulay (approved):
The producer-race blocker is addressed: disconnect wakeups now use the same send-buffer mutex as normal pushes, retaining the queue for the operation. Early subscription and watcher cleanup are preserved. I also added exception-safe cleanup to the active-production regression test in 4a3e8865.

Validation of 4a3e8865:
- Universal macOS Debug build passed; the focused [outlet],[open],[reopen],[sync] suite passed 1,710 assertions in 20 cases on both arm64 and x86_64.
- Injected assertion and exception failures confirmed the test stops and joins its pusher during unwinding.
- With Xcode 27 / Apple Clang 21 on native arm64, the same focused suite passed under ThreadSanitizer without warnings.
- 100 repeated disconnect-during-push runs passed all functional assertions, with no reports of the outlet queue producer race. Fifteen runs reported the separate sample-recycling warning now tracked in #305 and consequently exited unsuccessfully under TSan; this is not a claim of a fully TSan-clean stress run.
- All 14 PR CI checks passed.

Synchronous idle disconnect detection remains scoped to #303. Ready to merge this asynchronous-outlet fix.

PR #300, opened 2026-09-20
Ensure outlets are ready when open_stream returns

### LINKED ISSUE #176: lib.lsl_open_stream does not block until the stream is fully ready
I discovered this using the Python API, but it's most likely present everywhere. See below the desription and a MWE in Python:

In the Python API, `StreamInlet.open_stream(timeout)` is not completely blocking. If you create an outlet, create an inlet, open the stream on the inlet and then immediately push. a sample and pull a sample (with a positive timeout to wait until a sample appears in the inlet), it will hang out for the duration of the timeout specified in `pull_sample` because the sample was pushed before the stream was completely open. At least that's my interpretation. Here is a MWE:

```
import time
from pylsl import StreamInfo, StreamInlet, StreamOutlet, resolve_streams
sinfo = StreamInfo("test", "", 1, 0.0, "string", "test-stream")
outlet = StreamOutlet(sinfo)
stream = resolve_streams(wait_time=2)[0]
inlet = StreamInlet(stream)
inlet.open_stream(timeout=1000)
time.sleep(0.1)  # comment to show the issue
outlet.push_sample("1")
data, ts = inlet.pull_sample(timeout=2)
assert len(data) == 1
```

With the sleep, `data` is received; without the sleep, `inlet.pull_sample` exit after 2 seconds and `data` is `None` .

cboulay:
A sample pushed immediately after `open_stream()` returned could be lost: the inlet had already received the handshake, but the outlet's handshake-completion callback had not yet registered its consumer.

Register the asynchronous consumer queue before sending the feed header. The write callback owns the queue until it can pass it to the transfer thread; a failed handshake releases it. Sample transmission still starts only after the header write completes, preserving wire ordering and the existing zero-buffer handshake behavior.

For synchronous outlets, send the header under the existing sample-write lock and register the socket before releasing that lock. This prevents an immediate push from missing the socket or overtaking the handshake; the configured synchronous send timeout also applies to this write.

The regression test opens a fresh inlet and immediately pushes exactly one sample, with no sleeps, `wait_for_consumers()`, or retries. It covers 30 connections each for asynchronous numeric, synchronous numeric, and string streams.

Validation on macOS:

- The regression test fails against the pre-fix library on its first trial.
- All 90 connections pass with the fix (183 assertions).
- Universal x86_64/arm64 build succeeds, with benchmarks enabled.
- Internal suite passes: 34 cases pass, one existing case fails as expected (969 assertions total).
- Exported suite passes with benchmark cases excluded: 2,365 assertions across 35 cases.
- Windows and Linux execution will be checked by CI.

Based directly on `dev`; independent of #299. Closes #176.


cboulay:
@copilot resolve the merge conflicts in this pull request

